PR #133, opened 2017-04-08
[CLI] Set service name dynamically if not provided

eculver:
Previously, users had to provide `--service` flag for any non-production scope.

This also adds configuration for positional arguments to be required for `scope` subcommands and to be optional for `schema` subcommands.

Example output:
```
$ dosa scope create
Error: the required argument `scopes (at least 1 argument)` was not provided
```
```
$ dosa scope create -h
Error: Usage:
  dosa [OPTIONS] scope create scopes...
...
```
```
$ dosa schema check -h
Error: Usage:
  dosa [OPTIONS] schema [schema-OPTIONS] check [check-OPTIONS] [paths...]
...
```

### DIFF
--- a/cmd/dosa/main.go
+++ b/cmd/dosa/main.go
@@ -38,7 +38,7 @@ type GlobalOptions struct {
 	Host        string   `long:"host" default:"127.0.0.1" description:"The hostname or IP for the gateway."`
 	Port        string   `short:"p" long:"port" default:"21300" description:"The hostname or IP for the gateway."`
 	Transport   string   `long:"transport" default:"tchannel" description:"TCP Transport to use. Options: http, tchannel."`
-	ServiceName string   `long:"service" default:"dosa-gateway" description:"The TChannel service name for the gateway."`
+	ServiceName string   `long:"service" description:"The TChannel service name for the gateway."`
 	CallerName  string   `long:"caller" default:"dosacli-$USER" description:"Caller will override the default caller name (which is dosacli-$USER)."`
 	Timeout     timeFlag `long:"timeout" default:"60s" description:"The timeout for gateway requests. E.g., 100ms, 0.5s, 1s. If no unit is specified, milliseconds are assumed."`
 	Connector   string   `hidden:"true" long:"connector" default:"yarpc" description:"Name of connector to use"`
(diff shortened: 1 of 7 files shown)


cmd/dosa/schema.go
```diff
@@ -62,6 +62,15 @@ func (c *SchemaCmd) doSchemaOp(name string, f func(dosa.AdminClient, context.Con
 		fmt.Printf("options are %+v\n", *c)
 		fmt.Printf("global options are %+v\n", options)
 	}
+
+	// if not given, set the service name dynamically based on scope
+	if options.ServiceName == "" {
```
rkuris: nit: this is really a mapping of scope->service that should be easier to edit than this; recommend an array of globs to match against, with "*" as the final one:

```
type ScopeServiceMap struct {
    globPattern string
    serviceName string
} []scopeServiceMap = {
 { "production", "dosa-gateway" },
 { "*", "dosa-dev-gateway" },
}
```

cmd/dosa/main.go
```diff
@@ -38,7 +38,7 @@ type GlobalOptions struct {
 	Host        string   `long:"host" default:"127.0.0.1" description:"The hostname or IP for the gateway."`
 	Port        string   `short:"p" long:"port" default:"21300" description:"The hostname or IP for the gateway."`
 	Transport   string   `long:"transport" default:"tchannel" description:"TCP Transport to use. Options: http, tchannel."`
-	ServiceName string   `long:"service" default:"dosa-gateway" description:"The TChannel service name for the gateway."`
+	ServiceName string   `long:"service" description:"The TChannel service name for the gateway."`
```
rkuris: nit: we lost the default value information (I realize it is no longer a constant)

cmd/dosa/schema.go
```diff
@@ -99,27 +108,38 @@ func (c *SchemaCmd) doSchemaOp(name string, f func(dosa.AdminClient, context.Con
...
 	*SchemaCmd
+	Args struct {
+		Paths []string `positional-arg-name:"paths"`
+	} `positional-args:"yes"`
 }
 
 // Execute executes a schema check command
 func (c *SchemaCheck) Execute(args []string) error {
-	return c.doSchemaOp("schema check", dosa.AdminClient.CheckSchema, args)
+	fmt.Println("Paths")
```
rkuris: Remove this debug info

PR #134, opened 2017-04-12
Add status cli command

junchaowu:


cmd/dosa/schema.go
```diff
@@ -131,6 +131,53 @@ func (c *SchemaUpsert) Execute(args []string) error {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), options.Timeout.Duration())
+	defer cancel()
+
+	status, err := client.CheckSchemaStatus(ctx, c.NamePrefix, c.Version)
+	if err != nil {
+		if c.Verbose {
+			fmt.Printf("detail:%+v\n", err)
+		}
+		fmt.Println("Status: NOT OK")
```
zhenkai: Can we just print the error msg instead of "NOT OK"?

doc.go
```diff
@@ -22,7 +22,9 @@
 //
 // Abstract
 //
-// [dosa][dosa-spec] is a library that provides a distributed
+// DOSA open source is not ready for prime time yet. We will announce when it's ready.
```
zhenkai: why add this?

junchaowu: We have a build tool to make go doc based on readme. Readme has been updated to those statement.

cmd/dosa/schema.go
junchaowu: The `err` will return to the main(). main() will return error status code.

PR #138, opened 2017-04-28
Add support for complex type

SunZhaonan:
- Purpose:
Provide complex type support. When define schema entities, user can define their custom types. All they need to do is just implement **CustomObjectInterface** interface and provide marshal and unmarshal function. Then this field can be used as a type in entity object.

- Content of this diff:
1. Define a interface for **CustomObjectInterface**.
2. Doing Marshal and Unmarshal for yarpc call.
3. Schema management in cql, convert custom object to byte. 
4. Test

### DIFF
--- a/connectors/yarpc/helpers.go
+++ b/connectors/yarpc/helpers.go
@@ -54,12 +54,12 @@ func RawValueAsInterface(val dosarpc.RawValue, col dosa.ColumnDefinition) interf
 		if col.CustomType == nil {
 			panic("no custom type defined")
 		}
-		newObject := reflect.Zero(col.CustomType).Elem().Interface().(dosa.CustomObjectInterface)
-		err := newObject.Unmarshal(val.BinaryValue)
+		newObject := reflect.Zero(col.CustomType).Interface().(dosa.CustomObjectInterface)
+		result, err := newObject.Unmarshal(val.BinaryValue)
 		if err != nil {
 			panic(err)
 		}
-		return newObject
+		return result
 	}
 	panic("bad type")
 }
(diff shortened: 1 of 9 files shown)


rkuris:
I'm not convinced this is the right way to solve this problem. There are a few issues:

 * This isn't super easy to use and/or transparent. Each author must decide on their byte[] encoding, and having them inconsistent across all applications is a concern. I think we'd be better off just using JSON marshaling for the generic case, or allowing someone to specify JSON encoding for this, if we went with still keeping an underlying []byte type.
 * It's not clear from a schema service standpoint how someone handles additions to the embedded data type. That is, if I add a new string field, are the old ones compatible still? What about if we remove one? Does that mean that end users who set that field previously will have to handle the exception at runtime?
 * It might be better to pass this information down to the server for each field. This means that we could implement this with an underlying Cassandra UDT which can be managed easier.

junchaowu:
Thanks for writing this prototype. We need to think about more as Ron mention there are some other options. We will get you back soon.

SunZhaonan:
thanks for reply
- Marshaling : I'm OK with having a default encoding/decoding way inside DOSA. I was thinking about this option, the reason I didn't do it is because 1. user may want to customized marshal & unmarshal. 2. In this way the DOSA can restrict the type that can used for define field in entity. Only struct that implement this interface can used for define fields.
- Schema : I think we already have schema incompatible check for DOSA upsert schema. We can extend the schema service to check incompatible for embedded schema as well. Before this, we need to pass the embedded schema when register schema, which is not included in this pull request, but should be a easy fix.
- Cassandra UDT : Implementing with Cassandra UDT is the right way to go. That required more changes on gateway side.

So this proposal is basically adding a "syntax sugar" on client side, so the user can actually use a 'struct' to define their embedded schema. (instead of use byte[]). I think a lot of clients (almost all) will have this issue, so it make sense to have this functionality build in DOSA clients.

rkuris:
After some discussion, decided to close this PR.

PR #139, opened 2017-05-01
Add mock client

eculver:
This adds a mock client implementation generated by gomock and runnable examples for using it.


examples/testing/README.md
```diff
@@ -0,0 +1,70 @@
...
+        c2.EXPECT().Initialize(gomock.Any()).Return(nil).Times(1)
+        c2.EXPECT().Read(gomock.Any(), nil, user).Return(nil).Times(1)
+        ds2, _ := examples.NewDatastore(c2)
+
+        u2, err2 := ds2.GetUser(ctx, uuid)
+        assert.NoError(t, err2)
+        assert.Equal(t, u2, user)
+    }
+
+A complete, runnable example of this can be found in our [testing examples package](https://github.com/uber-go/dosa/tree/master/examples/testing).
```
rkuris: We should add instructions on how we made the mockgen file (what options were passed, etc) so it can be regenerated if the client should change in a later release.

eculver: ACK. Can just run `make mocks` now. I also added a README.

PR #141, opened 2017-05-03
Remove references to pborman UUID package

rkuris:


junchaowu:
Can you update the doc as well? It is kind of breaking change.

PR #144, opened 2017-05-12
Additional documentation for Client interface

rkuris:
Named everything.
Also added some notes about what you must fill out before
calling various methods

### DIFF
--- a/client.go
+++ b/client.go
@@ -92,15 +92,20 @@ func ErrorIsAlreadyExists(err error) bool {
 // Client defines the methods to operate with DOSA entities
 type Client interface {
 	// Initialize must be called before any data operation
-	Initialize(context.Context) error
+	Initialize(ctx context.Context) error
 
 	// Create creates an entity; it fails if the entity already exists.
+	// You must fill in all of the fields of the DomainObject before
+	// calling this method, or they will be inserted with the zero value
 	// This is a relatively expensive operation. Use Upsert whenever possible.
-	CreateIfNotExists(context.Context, DomainObject) error
+	CreateIfNotExists(ctx context.Context, objectToCreate DomainObject) error
 
 	// Read fetches a row by primary key. A list of fields to read can be
 	// specified. Use All() or nil for all fields.
-	Read(context.Context, []string, DomainObject) error
+	// Before calling this method, fill in the DomainObject with ALL
+	// of the primary key fields; the other field values will be populated
+	// as a result of the read
+	Read(ctx context.Context, fieldsToRead []string, objectToRead DomainObject) error
 
 	// TODO: Coming in v2.1
 	// MultiRead fetches several rows by primary key. A list of fields can be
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -92,15 +92,19 @@ func ErrorIsAlreadyExists(err error) bool {
...
+	// calling this method, or they will be inserted with the zero value
 	// This is a relatively expensive operation. Use Upsert whenever possible.
-	CreateIfNotExists(context.Context, DomainObject) error
+	CreateIfNotExists(ctx context.Context, objectToCreate DomainObject) error
 
 	// Read fetches a row by primary key. A list of fields to read can be
 	// specified. Use All() or nil for all fields.
-	Read(context.Context, []string, DomainObject) error
+	// Before calling this method, fill in the DomainObject with ALL
+	// of the primary key fields; the other field values will be returned
```
klnusbaum: Instead of:
> the other field values will be returned

Maybe:

> DOSA will populate the other fields as a result of the read

client.go
```diff
@@ -109,30 +113,40 @@ type Client interface {
```
klnusbaum: Do you think it's also worth updating the functions we currently have commented out?
There are a few others that we might also want to update as well in that case

client.go
```diff
@@ -109,30 +113,40 @@ type Client interface {
...
+	// Before calling range, create a RangeOp and fill in the table
+	// along with the partition key information. You will get back
+	// an array of DomainObjects, which will be of the type you requested
+	// in the rangeOp
+	Range(ctx context.Context, rangeOp *RangeOp) ([]DomainObject, string, error)
 
 	// Search fetches entities by fields that have been marked "searchable"
-	Search(context.Context, *SearchOp) ([]DomainObject, string, error)
+	// TODO: Coming in v2.1
+	// Search(ctx context.Context, *SearchOp) ([]DomainObject, string, error)
```
klnusbaum: is there a reason we named the context parameter but not the searchOp parameter?

client.go
```diff
@@ -109,30 +113,40 @@ type Client interface {
...
 	// contain the primary key field values.
 	// MultiRemove(context.Context, ...DomainObject) (MultiResult, error)
 
 	// Range fetches entities within a range
-	Range(context.Context, *RangeOp) ([]DomainObject, string, error)
+	// Before calling range, create a RangeOp and fill in the table
+	// along with the partition key information. You will get back
+	// an array of DomainObjects, which will be of the type you requested
+	// in the rangeOp
+	Range(ctx context.Context, rangeOp *RangeOp) ([]DomainObject, string, error)
```
klnusbaum: For this and Scan, we should probably also document what the return arguments mean, (at least what the string return argument is. What is that?)

client.go
```diff
@@ -109,30 +114,44 @@ type Client interface {
...
-	Search(context.Context, *SearchOp) ([]DomainObject, string, error)
+	// TODO: Coming in v2.1
+	// Search(ctx context.Context, searchOp *SearchOp) ([]DomainObject, string, error)
 
 	// ScanEverything fetches all entities of a type
-	ScanEverything(context.Context, *ScanOp) ([]DomainObject, string, error)
+	// Before calling ScanEverything, create a scanOp to specify the
+	// table to scan. The return values are an array of objects, that
+	// you can type-assert to the appropriate dosa.Entity, a string
+	// that contains the continuation token, and any error.
```
klnusbaum: This is a really good explination of the continuation token. I think it's worth adding up in the documentation of the `Range` as well.

PR #146, opened 2017-05-16
Add Matchers for RangeOp and ScanOp

klnusbaum:
I was going to write some unit tests for some code that was using DOSA, and I realized I didn't have a good way to tell the mock DOSA client that it should expect a Range call with a certain RangeOp. So I went ahead and made a gomock Matcher for the RangeOp (and the ScanOp as well). The one thing I couldn't really figure out how to do was to compare on the DomainObject of the ScanOp when matching ScanOps. I figure it's not too big of a deal, but if anyone knows how I would love to be enlightened.

### DIFF
--- a/range.go
+++ b/range.go
@@ -25,8 +25,8 @@ import (
 	"fmt"
 	"sort"
 
-	"github.com/pkg/errors"
 	"github.com/golang/mock/gomock"
+	"github.com/pkg/errors"
 )
 
 // RangeOp is used to specify constraints to Range calls
@@ -145,12 +145,8 @@ func convertRangeOpConditions(r *RangeOp, t *Table) (map[string][]*Condition, er
 	return serverConditions, nil
 }
 
-type CondsMantcher struct {}
-
-
-
 type rangeOpMatcher struct {
-	conds  map[Condition]bool
+	conds    map[Condition]bool
 	eqScanOp gomock.Matcher
 }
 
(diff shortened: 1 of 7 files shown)


range.go
```diff
@@ -143,3 +144,48 @@ func convertRangeOpConditions(r *RangeOp, t *Table) (map[string][]*Condition, er
...
+		}
+	}
+
+	return rangeOpMatcher{
+		conds: conds,
+		eqScanOp: EqScanOp(&(op.sop)),
+	}
+}
+func (m rangeOpMatcher) Matches(x interface{}) bool {
+	op := x.(*RangeOp)
```
rkuris: This panics if you don't pass in a *RangeOp. Can we do better?

scan.go
```diff
@@ -74,3 +78,55 @@ func (s *ScanOp) Fields(fields []string) *ScanOp {
...
+
+	return scanOpMatcher{
+		limit:  op.limit,
+		token:  op.token,
+		fields: fields,
+		typ:    reflect.TypeOf(op.object).Elem(),
+	}
+}
+
+func (m scanOpMatcher) Matches(x interface{}) bool {
```
eculver: nit: I think the linter should bark at you for this one, should have a docstring.

scan.go
```diff
@@ -74,3 +78,55 @@ func (s *ScanOp) Fields(fields []string) *ScanOp {
...
+	for _, field := range op.fieldsToRead {
+		if !m.fields[field] {
+			return false
+		}
+	}
+
+	return op.limit == m.limit && op.token == m.token && reflect.TypeOf(op.object).Elem() == m.typ
+}
+
+func (m scanOpMatcher) String() string {
```
eculver: nit: same here about docstring, for `String` I always say "String satisfies the Stringer interface"

PR #147, opened 2017-05-20
Add memory connector

rkuris:
Will not be merged until the TODO list size is smaller (perhaps 0)

Implemented: CreateIfNotExists, Read, Upsert, Delete
Partially implemented: Range, Scan

TODO:
Range and Scan always return the whole partition that matches.
As a result, they don't return or parse continuation tokens

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -74,13 +74,15 @@ func ExampleNewClient() {
 	reg, err := dosaRenamed.NewRegistrar("test", "myteam.myservice", cte1)
 	if err != nil {
 		// registration will fail if the object is tagged incorrectly
-		panic("dosaRenamed.NewRegister returned an error")
+		fmt.Printf("NewRegistrar error: %s", err)
+		return
 	}
 
 	// use a devnull connector for example purposes
 	conn, err := dosaRenamed.GetConnector("devnull", nil)
 	if err != nil {
-		panic(err)
+		fmt.Printf("GetConnector error: %s", err)
+		return
 	}
 
 	// create the client using the registry and connector
@@ -88,7 +90,8 @@ func ExampleNewClient() {
 
 	err = client.Initialize(context.Background())
 	if err != nil {
-		panic(err)
+		fmt.Printf("Initialize error: %s", err)
+		return
 	}
 }
 
@@ -101,7 +104,8 @@ func ExampleGetConnector() {
 	// your list of entities is important. In this case, we only have one, our ClientTestEntity1
 	reg, err := dosaRenamed.NewRegistrar("test", "myteam.myservice", &ClientTestEntity1{})
 	if err != nil {
-		panic(err)
+		fmt.Printf("NewRegistrar error: %s", err)
+		return
 	}
 
 	// Find the memory connector. There is no configuration information so pass a nil
@@ -114,7 +118,8 @@ func ExampleGetConnector() {
 
 	// initialize the client; this should always work for the in-memory connector
 	if err = client.Initialize(context.Background()); err != nil {
-		panic(err)
+		fmt.Printf("Initialize error: %s", err)
+		return
 	}
 
 	// now populate an entity and insert it into the memory store
@@ -122,7 +127,8 @@ func ExampleGetConnector() {
 		ID:    int64(1),
 		Name:  "rkuris",
 		Email: "rkuris@uber.com"}); err != nil {
-		panic(err)
+		fmt.Printf("CreateIfNotExists error: %s", err)
+		return
 	}
 
 	// create an entity to hold the read result, just populate the key
@@ -130,7 +136,8 @@ func ExampleGetConnector() {
 	// now read the data from the "database", all columns
 	err = client.Read(context.Background(), dosaRenamed.All(), &e)
 	if err != nil {
-		panic(err)
+		fmt.Printf("Read error: %s", err)
+		return
 	}
 	// great! It worked, so display the information we stored earlier
 	fmt.Printf("id:%d Name:%q Email:%q\n", e.ID, e.Name, e.Email)
--- a/client_test.go
+++ b/client_test.go
@@ -92,11 +92,11 @@ func ExampleNewClient() {
 	}
 }
 
-// ExampleGetConnector gets an in-memory connector that can be used for testing your code
+// ExampleGetConnector gets an in-memory connector that can be used for testing your code.
 // The in-memory connector always starts off with no rows, so you'll need to add rows to
 // your "database" before reading them
 func ExampleGetConnector() {
-	// register your entities so the engine can separate your data based on table names
+	// register your entities so the engine can separate your data based on table names.
 	// Scopes and prefixes are not used by the in-memory connector, and are ignored, but
 	// your list of entities is important. In this case, we only have one, our ClientTestEntity1
 	reg, err := dosaRenamed.NewRegistrar("test", "myteam.myservice", &ClientTestEntity1{})
@@ -208,7 +208,7 @@ func TestClient_Read(t *testing.T) {
 	assert.NoError(t, c3.Initialize(ctx))
 	assert.NoError(t, c3.Read(ctx, fieldsToRead, cte1))
 	assert.Equal(t, cte1.ID, results["id"])
-	assert.Equal(t, cte1.Name, results["name"])
+	assert.NotEqual(t, cte1.Name, results["name"])
 	assert.Equal(t, cte1.Email, results["email"])
 }
 
--- a/client_test.go
+++ b/client_test.go
@@ -802,4 +802,3 @@ func TestErrorIsAlreadyExists(t *testing.T) {
 	assert.True(t, dosaRenamed.ErrorIsAlreadyExists(errors.Wrap(&dosaRenamed.ErrAlreadyExists{}, "wrapped")))
 	assert.Equal(t, "already exists", (&dosaRenamed.ErrAlreadyExists{}).Error())
 }
-
--- a/client_test.go
+++ b/client_test.go
@@ -31,8 +31,11 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 
+	"fmt"
+
 	dosaRenamed "github.com/uber-go/dosa"
-	"github.com/uber-go/dosa/connectors/devnull"
+	_ "github.com/uber-go/dosa/connectors/devnull"
+	_ "github.com/uber-go/dosa/connectors/memory"
 	"github.com/uber-go/dosa/mocks"
 )
 
@@ -57,9 +60,15 @@ var (
 	ctx           = context.TODO()
 	scope         = "test"
 	namePrefix    = "team.service"
-	nullConnector = &devnull.Connector{}
+	nullConnector dosaRenamed.Connector
 )
 
+func init() {
+	nullConnector, _ = dosaRenamed.GetConnector("devnull", nil)
+}
+
+// ExampleNewClient initializes a client using the devnull connector, which discards all
+// the data you send it and always returns no rows. It's only useful for testing dosa.
 func ExampleNewClient() {
 	// initialize registrar
 	reg, err := dosaRenamed.NewRegistrar("test", "myteam.myservice", cte1)
@@ -69,15 +78,63 @@ func ExampleNewClient() {
 	}
 
 	// use a devnull connector for example purposes
-	conn := &devnull.Connector{}
+	conn, err := dosaRenamed.GetConnector("devnull", nil)
+	if err != nil {
+		panic(err)
+	}
 
 	// create the client using the registry and connector
 	client := dosaRenamed.NewClient(reg, conn)
 
 	err = client.Initialize(context.Background())
 	if err != nil {
-		errors.Wrap(err, "client.Initialize returned an error")
+		panic(err)
 	}
+}
+
+// ExampleGetConnector gets an in-memory connector that can be used for testing your code
+// The in-memory connector always starts off with no rows, so you'll need to add rows to
+// your "database" before reading them
+func ExampleGetConnector() {
+	// register your entities so the engine can separate your data based on table names
+	// Scopes and prefixes are not used by the in-memory connector, and are ignored, but
+	// your list of entities is important. In this case, we only have one, our ClientTestEntity1
+	reg, err := dosaRenamed.NewRegistrar("test", "myteam.myservice", &ClientTestEntity1{})
+	if err != nil {
+		panic(err)
+	}
+
+	// Find the memory connector. There is no configuration information so pass a nil
+	// For this to work, you must force the init method of memory to run first, which happens
+	// when we imported memory in the import list, with an underscore to just get the side effects
+	conn, _ := dosaRenamed.GetConnector("memory", nil)
+
+	// now construct a client from the registry and the connector
+	client := dosaRenamed.NewClient(reg, conn)
+
+	// initialize the client; this should always work for the in-memory connector
+	if err = client.Initialize(context.Background()); err != nil {
+		panic(err)
+	}
+
+	// now populate an entity and insert it into the memory store
+	if err := client.CreateIfNotExists(context.Background(), &ClientTestEntity1{
+		ID:    int64(1),
+		Name:  "rkuris",
+		Email: "rkuris@uber.com"}); err != nil {
+		panic(err)
+	}
+
+	// create an entity to hold the read result, just populate the key
+	e := ClientTestEntity1{ID: int64(1)}
+	// now read the data from the "database", all columns
+	err = client.Read(context.Background(), dosaRenamed.All(), &e)
+	if err != nil {
+		panic(err)
+	}
+	// great! It worked, so display the information we stored earlier
+	fmt.Printf("id:%d Name:%q Email:%q\n", e.ID, e.Name, e.Email)
+	// Output: id:1 Name:"rkuris" Email:"rkuris@uber.com"
 }
 
 func TestNewClient(t *testing.T) {
@@ -745,3 +802,4 @@ func TestErrorIsAlreadyExists(t *testing.T) {
 	assert.True(t, dosaRenamed.ErrorIsAlreadyExists(errors.Wrap(&dosaRenamed.ErrAlreadyExists{}, "wrapped")))
 	assert.Equal(t, "already exists", (&dosaRenamed.ErrAlreadyExists{}).Error())
 }
+
--- a/connector.go
+++ b/connector.go
@@ -92,6 +92,9 @@ type SchemaStatus struct {
 
 // Connector is the interface that must be implemented for a backend service
 // It can also be implemented using an RPC such as thrift (dosa-idl)
+// When fields are returned from read/range/search/scan methods, it's legal for the connector
+// to return more fields than originally requested. The caller of the connector should never mutate
+// the returned columns either, in case they are from a cache
 type Connector interface {
 	// DML operations (CRUD + search)
 	// CreateIfNotExists creates a row, but only if it does not exist.
--- a/connector.go
+++ b/connector.go
@@ -100,11 +100,11 @@ type Connector interface {
 	// CreateIfNotExists creates a row, but only if it does not exist.
 	CreateIfNotExists(ctx context.Context, ei *EntityInfo, values map[string]FieldValue) error
 	// Read fetches a row by primary key
-	// If fieldsToRead is empty or nil, all non-key fields would be fetched.
-	Read(ctx context.Context, ei *EntityInfo, keys map[string]FieldValue, fieldsToRead []string) (values map[string]FieldValue, err error)
+	// If minimumFields is empty or nil, all non-key fields would be fetched.
+	Read(ctx context.Context, ei *EntityInfo, keys map[string]FieldValue, minimumFields []string) (values map[string]FieldValue, err error)
 	// MultiRead fetches several rows by primary key
-	// If fieldsToRead is empty or nil, all non-key fields would be fetched.
-	MultiRead(ctx context.Context, ei *EntityInfo, keys []map[string]FieldValue, fieldsToRead []string) (results []*FieldValuesOrError, err error)
+	// If minimumFields is empty or nil, all non-key fields would be fetched.
+	MultiRead(ctx context.Context, ei *EntityInfo, keys []map[string]FieldValue, minimumFields []string) (results []*FieldValuesOrError, err error)
 	// Upsert updates some columns of a row, or creates a new one if it doesn't exist yet.
 	Upsert(ctx context.Context, ei *EntityInfo, values map[string]FieldValue) error
 	// MultiUpsert updates some columns of several rows, or creates a new ones if they doesn't exist yet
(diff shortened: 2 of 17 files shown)


rkuris:
This code becomes much simpler if https://github.com/satori/go.uuid/pull/50 gets accepted

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	"time"
+
+	"encoding/binary"
+
+	"github.com/satori/go.uuid"
+	"github.com/uber-go/dosa"
+	"github.com/uber-go/dosa/connectors/base"
+)
+
+// Connector is an in memory connector.
```
eculver: nit: typos
"in memory" -> "in-memory"
"it's" -> "its"

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+// CreateIfNotExists inserts a row if it isn't already there. The basic flow is:
+// Find the partition, if it's not there, then create it and insert the row there
+// If the partition is there, and there's data in it, and there's no clustering key, then fail
+// Otherwise, search the partition for the exact same clustering keys. If there, fail
+// if not, then insert it at the right spot (sort.Search does most of the heavy lifting here)
+func (c *Connector) CreateIfNotExists(_ context.Context, ei *dosa.EntityInfo, values map[string]dosa.FieldValue) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if c.data == nil {
+		c.data = make(map[string]map[string][]map[string]dosa.FieldValue)
```
eculver: this should be moved to a constructor so that it doesn't have to be repeated in each method.

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+	if c.data[ei.Def.Name] == nil {
```
eculver: nit: I think you want to check existence and use the value like this instead:
```
val, exists := c.data[ei.Def.Name]
if !exists {
    ...
}
```

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+		c.data[ei.Def.Name] = make(map[string][]map[string]dosa.FieldValue)
+	}
+	entityRef := c.data[ei.Def.Name]
+	encodedPartitionKey := partitionKeyBuilder(ei, values)
+	if entityRef[encodedPartitionKey] == nil {
+		entityRef[encodedPartitionKey] = make([]map[string]dosa.FieldValue, 0, 1)
```
eculver: you can omit the capacity arg when making maps (and usually the size too); the runtime will allocate enough space for the size and type provided.

From  [the docs](https://golang.org/pkg/builtin/#make): 
> An empty map is allocated with enough space to hold the
> specified number of elements. The size may be omitted, in which case
> a small starting size is allocated.
> 

EDIT: this is actually broken: https://play.golang.org/p/ijSYsEoalr

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+	partitionRef := entityRef[encodedPartitionKey]
+	// no data in this partition? easy out!
+	if len(partitionRef) == 0 {
```
eculver: Same as above, the base case can be checked for ahead of time:
```
entityRef, exists := c.data[ei.Def.Name]
if !exists {
    ...
}
```

connectors/memory/memory.go
rkuris: it's not, there is a 3-arg version when you use a slice, see https://golang.org/pkg/builtin/#make

I wanted to provide a hint, we know we will be adding 1 item but I want it empty to begin with

connectors/memory/memory.go
rkuris: actually, that won't work -- if the last entry is deleted, exists will be true, but the slice is empty

connectors/memory/memory.go
rkuris: Since this can be set to nil directly and can still exist, I prefer this check

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	return nil
+}
+
+// Read searches for a row. First, it finds the partition, then it searches in the partition for
+// the data, and returns it when it finds it. Again, sort.Search does most of the heavy lifting
+// within a partition
+func (c *Connector) Read(_ context.Context, ei *dosa.EntityInfo, values map[string]dosa.FieldValue, fieldsToRead []string) (map[string]dosa.FieldValue, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	if c.data[ei.Def.Name] == nil {
```
eculver: nit: "DRY" - suggest moving into `partitionKeyBuilder` or another helper

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+
+	if len(ei.Def.ClusteringKeySet()) == 0 {
+		return filterSet(partitionRef[0], fieldsToRead), nil
+	}
+	// clustering key, search for the value in the set
+	found, inx := findInsertionPoint(ei, partitionRef, values)
+	if !found {
+		return nil, &dosa.ErrNotFound{}
+	}
+	return filterSet(partitionRef[inx], fieldsToRead), nil
```
eculver: this should also return `&dosa.ErrNotFound{}` if `filterSet` returns 0/nil

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	// perform slice magic to insert value at given offset
+	l := len(entityRef[encodedPartitionKey])                                                                     // get length
+	entityRef[encodedPartitionKey] = append(entityRef[encodedPartitionKey], entityRef[encodedPartitionKey][l-1]) // copy last element
+	// scoot over remaining elements
+	copy(entityRef[encodedPartitionKey][offset+1:], entityRef[encodedPartitionKey][offset:])
+	// and plunk value into appropriate location
+	entityRef[encodedPartitionKey][offset] = values
+	return nil
+}
+func merge(into map[string]dosa.FieldValue, from map[string]dosa.FieldValue) {
```
eculver: nit: not needed if `sort.Sort` called on every insert.

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+func (c *Connector) Range(_ context.Context, ei *dosa.EntityInfo, columnConditions map[string][]*dosa.Condition, fieldsToRead []string, token string, limit int) ([]map[string]dosa.FieldValue, string, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	if c.data[ei.Def.Name] == nil {
+		return nil, "", &dosa.ErrNotFound{}
+	}
+	entityRef := c.data[ei.Def.Name]
+
+	// find the equals conditions on each of the partition keys
+	values := map[string]dosa.FieldValue{}
```
eculver: nit: use `make` for consistency (and efficiency): `values := make(map[string]dosa.FieldValue)`

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	for _, pk := range ei.Def.Key.PartitionKeys {
+		// TODO: assert len(columnConditions[pk] == 1
+		// TODO: assert columnConditions[pk][0].Op is equals
+		values[pk] = columnConditions[pk][0].Value
+	}
+
```
eculver: missing a check to ensure that all partition keys are accounted for, can bail early if missing.

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	encodedPartitionKey := partitionKeyBuilder(ei, values)
+	partitionRef := entityRef[encodedPartitionKey]
+	// no data in this partition? easy out!
+	if len(partitionRef) == 0 {
+		return nil, "", &dosa.ErrNotFound{}
+	}
+	// hunt through the partitionRef and return values that match search criteria
+	// TODO: This can be done much faster using a binary search
+	startinx, endinx := 0, len(partitionRef)-1
+	for startinx < len(partitionRef) && !matchesClusteringConditions(ei, columnConditions, partitionRef[startinx]) {
```
eculver: you don't need to bother searching against all columns if the data is already sorted; also, this has the side-effect of falsely returning matches where clustering key value is the 0 value

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+	// TODO: adjust startinx with a passed in token
+	for endinx >= startinx && !matchesClusteringConditions(ei, columnConditions, partitionRef[endinx]) {
+		endinx--
+
+	}
+	if endinx <= startinx {
+		return nil, "", &dosa.ErrNotFound{}
+	}
+	// TODO: enforce limits and return a token when there are more rows
```
eculver: this is just a comparison of whether the length of the "filtered" slice is:
1) <= length of entire row
2) > max allowed

correct?

connectors/memory/memory.go
eculver: if there is no value for key `ei.Def.Name`, `exists == false` -- `c.data` is a map

connectors/memory/memory_test.go
```diff
@@ -0,0 +1,650 @@
...
+	assert.Equal(t, dosa.FieldValue(int64(1)), vals["c1"])
+}
+
+func TestConnector_Read(t *testing.T) {
+	sut := Connector{}
+
+	// read with no data
+	vals, err := sut.Read(context.TODO(), testEi, map[string]dosa.FieldValue{
+		"f1": dosa.FieldValue("data")}, []string{"c1"})
+	assert.Error(t, err)
```
klnusbaum: unecessary, the asset below assures this.

connectors/memory/memory_test.go
```diff
@@ -0,0 +1,650 @@
...
+		"c2": dosa.FieldValue(float64(2)),
+	})
+	assert.NoError(t, err)
+
+	vals, err = sut.Read(context.TODO(), testEi, map[string]dosa.FieldValue{
+		"f1": dosa.FieldValue("data")}, []string{"c1"})
+	assert.NoError(t, err)
+	assert.Equal(t, int64(1), vals["c1"])
+	assert.Nil(t, vals["c2"])
+	assert.Equal(t, 1, len(vals))
```
klnusbaum: nit: the assert package allows you to semantically assert length like so:
```
assert.Len(t, vals, 1)
```

Save you a few chars

connectors/memory/memory_test.go
```diff
@@ -0,0 +1,650 @@
...
+
+	vals, err = sut.Read(context.TODO(), testEi, map[string]dosa.FieldValue{
+		"f1": dosa.FieldValue("data")}, dosa.All())
+	assert.NoError(t, err)
+	assert.Equal(t, int64(1), vals["c1"])
+	assert.Equal(t, float64(2), vals["c2"])
+	assert.Equal(t, "data", vals["f1"])
```
klnusbaum: also, add an assert vals is of length 3

client_test.go
```diff
@@ -31,8 +31,11 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 
+	"fmt"
+
 	dosaRenamed "github.com/uber-go/dosa"
-	"github.com/uber-go/dosa/connectors/devnull"
+	_ "github.com/uber-go/dosa/connectors/devnull"
```
rkuris: Because this has the side effect of calling the init() method in the devnull package, if it hasn't already been called, which registers the devnull connector so it can be fetched by name later.

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+// writes are not. There is no attempt to improve the concurrency of the read or write path by
+// adding more granular locks.
+type Connector struct {
+	base.Connector
+	data map[string]map[string][]map[string]dosa.FieldValue
+	lock sync.RWMutex
+}
+
+// partitionKeyBuilder extracts the partition key components from the map and encodes them,
+// generating a unique string. It uses the encoding/gob method to make a byte array as the
```
rkuris: I can't hash the value, because it can collide. This has to produce a unique string given a set of inputs

client_test.go
```diff
@@ -69,15 +78,63 @@ func ExampleNewClient() {
...
+	// Scopes and prefixes are not used by the in-memory connector, and are ignored, but
+	// your list of entities is important. In this case, we only have one, our ClientTestEntity1
+	reg, err := dosaRenamed.NewRegistrar("test", "myteam.myservice", &ClientTestEntity1{})
+	if err != nil {
+		panic(err)
+	}
+
+	// Find the memory connector. There is no configuration information so pass a nil
+	// For this to work, you must force the init method of memory to run first, which happens
+	// when we imported memory in the import list, with an underscore to just get the side effects
```
rkuris: This is a really common paradigm. See http://go-database-sql.org/importing.html for some explanations.

connectors/memory/memory.go
rkuris: I think I can actually pull this out into a common function that just takes a merge function that can return an error

client_test.go
klnusbaum: Apparently there is an `init()` function in the devnull package. I missed that. woops

client_test.go
klnusbaum: Fair enough. I disagree with this pattern. But that's a discussion for another day ;)

connectors/memory/memory.go
rkuris: I really do want this code to look at len(partitionRef) not check for existence. It might exist but it's an empty map. See this code for more details.

https://play.golang.org/p/Oo8ZtQGIN3

client_test.go
```diff
@@ -57,9 +60,15 @@ var (
 	ctx           = context.TODO()
 	scope         = "test"
 	namePrefix    = "team.service"
-	nullConnector = &devnull.Connector{}
+	nullConnector dosaRenamed.Connector
 )
 
+func init() {
+	nullConnector, _ = dosaRenamed.GetConnector("devnull", nil)
```
rkuris: It's existed as part of the client_test.go for a while. I just moved it to an init block and fetching it from the registered constructor instead of just getting an empty `devnull.Connector{}`. The `nullConnector` is used in several tests, such as `TestNewClient`.

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+func timeFromUUID(u uuid.UUID) int64 {
+	low := int64(binary.BigEndian.Uint32(u[0:4]))
+	mid := int64(binary.BigEndian.Uint16(u[4:6]))
+	hi := int64((binary.BigEndian.Uint16(u[6:8]) & 0x0fff))
+	return low + (mid << 32) + (hi << 48)
+}
+
+// compareType compares a single DOSA field based on the type. This code assumes the types of each
+// of the columns are the same, or it will panic
+func compareType(d1 dosa.FieldValue, d2 dosa.FieldValue) int8 {
```
rkuris: I'm going to skip these for now. This added quite a few lines of code and didn't seem to buy me anything.

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+
+// CreateIfNotExists inserts a row if it isn't already there. The basic flow is:
+// Find the partition, if it's not there, then create it and insert the row there
+// If the partition is there, and there's data in it, and there's no clustering key, then fail
+// Otherwise, search the partition for the exact same clustering keys. If there, fail
+// if not, then insert it at the right spot (sort.Search does most of the heavy lifting here)
+func (c *Connector) CreateIfNotExists(_ context.Context, ei *dosa.EntityInfo, values map[string]dosa.FieldValue) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if c.data == nil {
```
rkuris: Completed this in a different way by merging CreateIfNotExists and Upsert into a single method, mergedInsert, which takes a merging function.

connectors/memory/memory.go
rkuris: I improved this some, but didn't fix every instance of it.

connectors/memory/memory.go
rkuris: I don't think that's possible. filterSet only filters fields. I guess it's possible to return no fields, but there's still a row there, and so it should do the right thing

connectors/memory/memory.go
rkuris: As discussed, we shouldn't need to do this

connectors/memory/memory.go
```diff
@@ -0,0 +1,493 @@
...
+func (c *Connector) Remove(_ context.Context, ei *dosa.EntityInfo, values map[string]dosa.FieldValue) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if c.data[ei.Def.Name] == nil {
+		return &dosa.ErrNotFound{}
+	}
+	entityRef := c.data[ei.Def.Name]
+	encodedPartitionKey := partitionKeyBuilder(ei, values)
+	if entityRef[encodedPartitionKey] == nil {
+		return &dosa.ErrNotFound{}
```
rkuris: This is indeed a bug and has been fixed.

connectors/memory/memory.go
rkuris: Connectors can safely assume their inputs have been sanitized, per our early design discussions. This is because we are controlling the callers. If we want to have a sanitizing connector, we can do that too, but adding lots of checks at every layer isn't necessary.

connectors/memory/memory.go
rkuris: Yes, it's possible to improve this, but it's not easy. There is a TODO a few lines above this.

connectors/memory/memory.go
rkuris: No, we need to construct a token so this can be continued. The token needs to be a serialized `[]Conditions`, created based on the contents of the last row returned.

This is a tricky subject that needs a ton of documentation once it's written. Let's get this merged first and then address it.

connectors/memory/memory_test.go
```diff
@@ -0,0 +1,650 @@
...
+	err := sut.CreateIfNotExists(context.TODO(), testEi, map[string]dosa.FieldValue{
+		"f1": dosa.FieldValue("data"),
+	})
+	assert.NoError(t, err)
+
+	err = sut.CreateIfNotExists(context.TODO(), testEi, map[string]dosa.FieldValue{
+		"f1": dosa.FieldValue("data"),
+	})
+
+	assert.Error(t, err)
```
rkuris: fixed all instances of this

connectors/memory/memory_test.go
rkuris: fixed all instances of this

connectors/memory/memory_test.go
rkuris: Actually, the contract about which fields should be returned wasn't correct; I amended this change to correct and document it

client_test.go
```diff
@@ -69,15 +78,63 @@ func ExampleNewClient() {
 	}
 
 	// use a devnull connector for example purposes
-	conn := &devnull.Connector{}
+	conn, err := dosaRenamed.GetConnector("devnull", nil)
+	if err != nil {
+		panic(err)
```
eculver: nit: suggest returning `err` instead of `panic` since user may copy/paste this code; I think @junchaow suggested this when `ExampleNewClient` was first implemented.

client_test.go
eculver: Can we "ear-mark" this for discussion? I don't want to shave any more yaks, but we should be consistent, meaning either provide constructor methods or use this paradigm consistently throughout all connectors.

client_test.go
```diff
@@ -151,7 +208,7 @@ func TestClient_Read(t *testing.T) {
 	assert.NoError(t, c3.Initialize(ctx))
 	assert.NoError(t, c3.Read(ctx, fieldsToRead, cte1))
 	assert.Equal(t, cte1.ID, results["id"])
-	assert.Equal(t, cte1.Name, results["name"])
+	assert.NotEqual(t, cte1.Name, results["name"])
```
eculver: Is this a bugfix? If so, shouldn't it be an empty string?

connectors/memory/memory.go
```diff
@@ -0,0 +1,449 @@
...
+		allTheThings = append(allTheThings, vals...)
+	}
+	if len(allTheThings) == 0 {
+		return nil, "", &dosa.ErrNotFound{}
+	}
+	return allTheThings, "", nil
+}
+
+// CheckSchema is just a stub; there is no schema management for the in memory connector
+// since creating a new one leaves you with no data!
```
eculver: I love the enthusiastic docstrings 👍

client_test.go
rkuris: It's impossible for the devnull connector to return any errors, but I'll change it anyway.

client_test.go
rkuris: Actually, it's due to the changed contract as in the commit comments. In this case, the "name" field wasn't requested, but the connector returned it anyway, which changed the value of the field. One could argue this was a test bug before, or at a minimum the contract between the connector and the client wasn't well defined.

connectors/memory/memory.go
rkuris: Cool, can you approve the changes then so once I fix the panic I can merge it?

connector.go
```diff
@@ -92,6 +92,9 @@ type SchemaStatus struct {
 
 // Connector is the interface that must be implemented for a backend service
 // It can also be implemented using an RPC such as thrift (dosa-idl)
+// When fields are returned from read/range/search/scan methods, it's legal for the connector
+// to return more fields than originally requested. The caller of the connector should never mutate
+// the returned columns either, in case they are from a cache
```
klnusbaum: This feels really brittle to me. I can't quite articulate why though, so I won't block.

That said, if we're really going to make this change to the Connector contract, I would advise we change the name of the `fieldsToRead` parameter in read/range/search/scan functions. It's not obvious in those functions that parameter `fieldToRead` actually means "please get at least these fields but it's also cool if you also give me additional fields". Perhaps rename them to `requiredFields`?

connectors/memory/memory.go
```diff
@@ -486,8 +435,15 @@ func (c *Connector) Shutdown() error {
 	return nil
 }
 
+// NewConnector creates a new in-memory connector
+func NewConnector() *Connector {
```
klnusbaum: Beautiful. I really like this. One centralized place for initializing the Connector.

Quick Question, instead of returning a `*Connector`, would it make sense to just return a `dosa.Connector`. That way the struct isn't tightly coupled?

client_test.go
rkuris: We should do this consistently throughout all connectors. This design pattern is considered best practice for golang.

connector.go
rkuris: Good call, I like `minimumFields` since "required" has a different connotation.

connectors/memory/memory.go
rkuris: One interesting problem here is that in the future I'd like to allow someone to call a method in the memory connector that isn't part of the normal Connector interface -- for example, to clear or allocate a request level cache, LRU expire some items, etc. In that case, if you need the concrete implementation, you would want this method.

Arguably, as of right now, nobody should be calling this except perhaps tests, so it could be unexported. The correct way to get one of these is to call dosa.GetConnector.

connectors/memory/memory.go
eculver: 👍

connector.go
klnusbaum: `minimumFields` or maybe just the shorter `minFields` was actually something I had considered as well. I'd be happy with that :)

klnusbaum:
If you want to save some chars, you could shorten minimumFields to minFields. Other than that, this is good to go as far as I'm concerned :)

SunZhaonan:
wow nice!

PR #151, opened 2017-05-30
Add support for UberFx

eculver:
This adds a new package for providing a DOSA client that can be used as an UberFx module. 

TODO:
- client tests + YARPC (test recorder problems)
- registry test coverage

### DIFF
--- a/fx/client.go
+++ b/fx/client.go
@@ -18,39 +18,31 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// Package dosa is the DOSA - Declarative Object Storage Abstraction.
-//
-// Abstract
-//
-// :warning: DOSA is BETA software. It is not recommended for production use.
-// We will announce when it's ready.
-//
-//
-// DOSA (https://github.com/uber-go/dosa/wiki) is a storage framework that
-// provides a
-// delcarative object storage abstraction for applications in Golang
-// and (eventually) Java. DOSA is designed to relieve common headaches developers
-// face while building stateful, database-dependent services.
-//
-//
-// If you'd like to start by writing a small DOSA-enabled program, check out
-// the getting started guide (https://github.com/uber-go/dosa/wiki/Getting-Started-Guide).
-//
-// Overview
-//
-// DOSA is a storage library that supports:
-//
-// • methods to store and retrieve go structs
-//
-// • struct annotations to describe queries against data
-//
-// • tools to create and/or migrate database schemas
-//
-// • implementations that serialize requests to remote stateless servers
-//
-// Annotations
-//
-// This project is released under the MIT License (LICENSE.txt).
-//
-//
-package dosa
+package dosafx
+
+import (
+	"go.uber.org/fx/service"
+	"go.uber.org/yarpc"
+
+	"github.com/pkg/errors"
+	"github.com/uber-go/dosa"
+	"github.com/uber-go/dosa/client"
+	"github.com/uber-go/dosa/config"
+)
+
+// New creates a new DOSA client that can be registered as an FX module.
+func New(d *yarpc.Dispatcher, h service.Host) (dosa.Client, error) {
+	cfg, err := config.NewConfigForService(h)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not populate DOSA configuration")
+	}
+
+	// before the connection can be used, it must be started
+	cc := d.ClientConfig(cfg.Service())
+	if err := cc.GetUnaryOutbound().Start(); err != nil {
+		return nil, errors.Wrap(err, "could not start outbound connection")
+	}
+
+	// create client
+	return dosaclient.New(cfg)
+}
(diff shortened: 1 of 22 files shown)


fx/client.go
```diff
@@ -0,0 +1,77 @@
...
+	"go.uber.org/fx/service"
+	"go.uber.org/yarpc"
+
+	"github.com/pkg/errors"
+	"github.com/uber-go/dosa"
+	dosarpc "github.com/uber-go/dosa/connectors/yarpc"
+)
+
+// New creates a new DOSA client that can be registered as an FX module.
+func New(d *yarpc.Dispatcher, h service.Host, entities ...dosa.DomainObject) (dosa.Client, error) {
```
klnusbaum: This feels like it's going to be a little confusing to users. They're going to see this `entities` parameter on the end of the function and be like "Wait, I thought I already specified my entity paths in the config file. Would there be any major downside to getting rid of the entities parameter? What's the use case for it?

Also, by not having it, we can simplify some of the code below.

fx/registry.go
```diff
@@ -0,0 +1,105 @@
...
+	return re, nil
+}
+
+// FindAll returns all registered entities from its internal index.
+func (r *fxRegistrar) FindAll() ([]*dosa.RegisteredEntity, error) {
+	res := []*dosa.RegisteredEntity{}
+	for _, re := range r.idx {
+		res = append(res, re)
+	}
+	if len(res) == 0 {
```
klnusbaum: Before creating the slice and doing the for loop, I would actually just put this check before we do anything in this function:
```
if len(r.idx) == 0 {
// return error
}
```

fx/registry.go
```diff
@@ -0,0 +1,105 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package dosafx
```
klnusbaum: per our discussion, none of this code is FX specific. In fact, it's just a better version of what should be in the Dosa codebase anyway. Let's move this into to dosa and deprecate our current NewRegistrar function.

eculver: 👍

fx/client.go
dmcaulay: what's the vision for usage? how will the entities be registered?

klnusbaum: @eculver and I had a long conversation about this yesterday. We think we've come up with a better overall strategy for creating new clients that can obviate the need for users to even understand the entire entity registration process. @eculver is going to do a little work on that and see if we can't just obviate the need for most of this code.

cmd/dosa/main_test.go
```diff
@@ -57,11 +57,10 @@ func TestHostOptionButNothingElse(t *testing.T) {
 func TestInvalidHost(t *testing.T) {
 	c := StartCapture()
 	exit = func(r int) {}
-	os.Args = []string{"dosa", "--host", "invalid-hostname.", "schema", "check", "--prefix", "foo", "../../testentity"}
+	os.Args = []string{"dosa", "--host", "", "schema", "check", "--prefix", "foo", "../../testentity"}
```
eculver: An update to YARPC removed this check. Our code only checks to see if it's an empty string.

rkuris: The invalid hostname check is a different check than an empty hostname. We're more likely to see an invalid hostname than an empty one, and an invalid hostname should return an error that explains what is wrong very clearly.

Removing this check means we aren't returning a good error when the hostname is incorrect. Is this an upstream bug with yarpc, or are we interpreting the error from yarpc incorrectly now?

eculver: We'd have to add code that checks the validity of the hostname because that code no longer exists in YARPC. This is just a side-effect of a dependency update.

fx/client_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		OutboundMiddleware: yarpc.OutboundMiddleware{
+			Unary: recorder.NewRecorder(t),
+		},
+	}
+	return yarpc.NewDispatcher(ycfg)
+}
+
+func TestFx(t *testing.T) {
+	/* if the YARPC recorder wasn't pathological, these would work with
```
eculver: I'm still working on getting this to work as expected...

cmd/dosa/main_test.go
rkuris: Since a typo in a hostname is pretty common, we need to finish this before releasing this change.

eculver:
This has grown unwieldy, going to split it into smaller PRs

PR #152, opened 2017-05-30
Introducing NullInt type.

umerazad:
Initial change for Nullable types. Sending it early to get feedback about the API. bool/string/float64/time types to follow along with more tests.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -27,7 +27,8 @@
 //
 //
 // DOSA (https://github.com/uber-go/dosa/wiki) is a storage framework that
-// provides a *delcarative object storage abstraction* for applications in Golang
+// provides a
+// *delcarative object storage abstraction* for applications in Golang
 // and (eventually) Java. DOSA is designed to relieve common headaches developers
 // face while building stateful, database-dependent services.
 //
--- a/doc.go
+++ b/doc.go
@@ -27,8 +27,7 @@
 //
 //
 // DOSA (https://github.com/uber-go/dosa/wiki) is a storage framework that
-// provides a
-// *delcarative object storage abstraction* for applications in Golang
+// provides a *delcarative object storage abstraction* for applications in Golang
 // and (eventually) Java. DOSA is designed to relieve common headaches developers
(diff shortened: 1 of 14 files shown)


null_int.go
```diff
@@ -0,0 +1,103 @@
+package dosa
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strconv"
+)
+
+// NullInt is a type that holds an optional int64 value.
```
rkuris: I think we can use https://golang.org/src/database/sql/sql.go?s=4941:5029#L187 instead, which gives us all a lot more for free: Scan(), Value(), etc.

doc.go
```diff
@@ -22,17 +22,19 @@
 //
 // Abstract
 //
-// DOSA open source is not ready for prime time yet. We will announce when it's ready.
+// :warning: DOSA is *BETA* software. It is not recommended for production use.
```
rkuris: Not sure why this file is modified, looks like part of something else

null_int.go
umerazad: I didn't want to take dependency on the whole sql package just to bring in the Null types. I can add Scan() support if you'd like.

doc.go
umerazad: Make generates it on my box.

null_int.go
rkuris: database/sql is part of the go core, and we're only pulling in the types, so it's not a big deal. However, if you still feel strongly about making new types, I'm okay with that too.

I don't know that we need Scan, so if you want to make your own types please make the structures as close to the ones in database/sql as possible (same names and methods).

Otherwise looks good.

doc.go
rkuris: This might already be fixed; please rebase and see.

null_int.go
umerazad: I've embedded sql.NullInt64 in our type.

rkuris: Can we change this to NullInt64 instead of just NullInt? Otherwise LGTM.

PR #158, opened 2017-05-31
add RemoveRange for in-memory connector

klnusbaum:
We'd like to add the ability to remove a range of values from DOSA. This adds that ability to the in-memory connector.

### DIFF
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -35,14 +35,6 @@ import (
 	"github.com/uber-go/dosa/connectors/base"
 )
 
-// partitionRange represents one section of a partition.
-type partitionRange struct {
-	entityRef map[string][]map[string]dosa.FieldValue
-	partitionKey string
-	start int
-	end int
-}
-
 // Connector is an in-memory connector.
 // The in-memory connector stores its data like this:
 // map[string]map[string][]map[string]dosa.FieldValue
@@ -61,6 +53,30 @@ type Connector struct {
 	data map[string]map[string][]map[string]dosa.FieldValue
 	lock sync.RWMutex
 }
+
+// partitionRange represents one section of a partition.
+type partitionRange struct {
+	entityRef map[string][]map[string]dosa.FieldValue
+	partitionKey string
+	start int
+	end int
+}
+
+// delete deletes the values referenced by the partitionRange. This method can't be called more than
+// once. Calling it more than once will cause a panic.
+func (pr *partitionRange) delete() {
+	partitionRef := pr.entityRef[pr.partitionKey]
+	pr.entityRef[pr.partitionKey] = append(partitionRef[:pr.start], partitionRef[pr.end+1:]...)
+	pr.entityRef = nil
+	pr.partitionKey = ""
+	pr.start = 0
+	pr.end = 0
+}
+
+func (pr *partitionRange) values() []map[string]dosa.FieldValue {
+	return pr.entityRef[pr.partitionKey][pr.start : pr.end+1]
+}
+
 
 // partitionKeyBuilder extracts the partition key components from the map and encodes them,
 // generating a unique string. It uses the encoding/gob method to make a byte array as the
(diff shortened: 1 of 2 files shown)


connectors/memory/memory.go
```diff
@@ -326,13 +334,51 @@ func (c *Connector) Remove(_ context.Context, ei *dosa.EntityInfo, values map[st
...
+func (c *Connector) RemoveRange(_ context.Context, ei *dosa.EntityInfo, columnConditions map[string][]*dosa.Condition) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	partitionRange := c.findRange(ei, columnConditions)
+	if partitionRange == nil {
+		return nil
+	}
+
+	partitionRef := partitionRange.entityRef[partitionRange.partitionKey]
```
klnusbaum: This all feels really bad, would love a better way to do this with out causing
```
entityRef[partitionKey] = append(partitionRef[:start], partitionRef[end+1:]...)
```
to be really long

rkuris: Did you look at making methods on partitionRange that do the heavy lifting? Something like:
```
func (p *partitionRange) deleteRange(er *[]map[string]dosa.FieldValue) {
    er = append(er[:p.start], er[p.end:]...)
```
The caller would then do:
```
partitionRange.deleteRange(&entityRef[partitionKey])
```

connectors/memory/memory_test.go
```diff
@@ -239,6 +239,53 @@ func TestConnector_Remove(t *testing.T) {
...
+	const idcount = 10
+	sut := NewConnector()
+
+	// remove with no data
+	err := sut.RemoveRange(context.TODO(), clusteredEi, map[string][]*dosa.Condition{
+		"f1": {{Op: dosa.Eq, Value: dosa.FieldValue("data")}},
+	})
+	assert.NoError(t, err)
+
+	// insert some data into data/i/uuid with a random set of uuids
```
rkuris: nit: copy/paste from another test ;)

connectors/memory/memory.go
rkuris: If you don't like the pointer approach, you can return the updated value with the range deleted (I think the pointer approach is more elegant though)

connectors/memory/memory_test.go
```diff
@@ -239,6 +239,53 @@ func TestConnector_Remove(t *testing.T) {
...
+	for i, x := range data {
+		assert.Equal(t, x["c1"], int64(i))
+	}
+
+	// ensure all the uuids greater than the uuid at the midpoint are removed.
+	data, _, err = sut.Range(context.TODO(), clusteredEi, map[string][]*dosa.Condition{
+		"f1": {{Op: dosa.Eq, Value: dosa.FieldValue("data")}},
+		"c1": {{Op: dosa.Gt, Value: dosa.FieldValue(int64(4))}},
+	}, dosa.All(), "", 200)
+	t.Logf("Length of data is: %d\n", len(data))
```
rkuris: remove this debug code

connectors/memory/memory_test.go
```diff
@@ -239,6 +239,53 @@ func TestConnector_Remove(t *testing.T) {
...
+	assert.Error(t, err)
+	assert.True(t, dosa.ErrorIsNotFound(err))
+}
```
rkuris: missing test: remove everything from a partitionKey

connectors/memory/memory.go
```diff
@@ -326,13 +350,43 @@ func (c *Connector) Remove(_ context.Context, ei *dosa.EntityInfo, values map[st
...
 }
 
+// RemoveRange removes all of the elements in the range specified by the entity info and the column conditions.
+func (c *Connector) RemoveRange(_ context.Context, ei *dosa.EntityInfo, columnConditions map[string][]*dosa.Condition) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	partitionRange := c.findRange(ei, columnConditions)
+	if partitionRange == nil {
+		return nil
```
rkuris: Nit: complexity: if partitionRange!=nil { pr.delete()}

PR #162, opened 2017-06-01
Add support for entity discovery

eculver:
This adds a new `registry` package which implements support for "auto-discovery" of DOSA entities. It was broken out of #151. This will be used in subsequent changes to add support for UberFx.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,26 +35,26 @@ const (
 	_defExclude    = "_test.go"
 
 	// default timeouts in milliseconds
-	_defInitializeTimeout        = 10000
-	_defCreateIfNotExistsTimeout = 10000
-	_defReadTimeout              = 10000
-	_defUpsertTimeout            = 10000
-	_defRemoveTimeout            = 10000
-	_defRangeTimeout             = 10000
-	_defSearchTimeout            = 10000
-	_defScanEverythingTimeout    = 10000
+	_defCreateIfNotExistsTimeout = time.Duration(10 * time.Second)
+	_defInitializeTimeout        = time.Duration(10 * time.Second)
+	_defRangeTimeout             = time.Duration(10 * time.Second)
+	_defReadTimeout              = time.Duration(2 * time.Second)
+	_defRemoveTimeout            = time.Duration(2 * time.Second)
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -21,15 +21,51 @@
...
 type Config struct {
-	Scope      string       `yaml:"scope"`
-	NamePrefix string       `yaml:"namePrefix"`
-	Yarpc      yarpc.Config `yaml:"yarpc"`
+	Scope       string         `yaml:"scope"`
+	NamePrefix  string         `yaml:"namePrefix"`
+	Connector   string         `yaml:"connector"`
+	EntityPaths []string       `yaml:"entityPaths"`
+	Excludes    []string       `yaml:"excludes"`
+	Yarpc       yarpc.Config   `yaml:"yarpc"`
```
klnusbaum: This makes me a little nervous. Why is connector specific configuration being placed in the main configuration for DOSA? It's like the Yarpc connector is special for some reason.

config/config.go
```diff
@@ -21,15 +21,51 @@
...
 
 import (
+	"time"
+
 	"github.com/uber-go/dosa"
 	"github.com/uber-go/dosa/connectors/yarpc"
 )
 
+const (
+	_defaultConn = "yarpc"
```
eculver: 👍

config/config.go
eculver: It is is "special" in that it requires configuration whereas all other connectors do not require configuration. The config itself is in the `yarpc` package.

Also, if we remove it, we'll break backwards compatibility.

:shrug:

klnusbaum: sad times

klnusbaum: we should deprecate this field then. And the Yarpc connector should be able to just configure itself. That's beyond the scope of this PR though. When you merge this, I'll get to work on it.

rkuris: IMHO the right way to handle this is to make the configuration a set of K/V pairs. The connector supports having arbitrary K/V pairs.

A second possible implementation is to just pass the yaml parts down to the connector. The attribute map has {}interface as the value anyway, so you can just not parse it and pass it downstream.

klnusbaum: Yea, I think the second bit is the way to go.

PR #163, opened 2017-06-01
Add dosaclient package for supporting auto-discovery of entities

eculver:
This implements the new registry package from #162 to both dynamically select connector based on configuration value and to "auto-discover" DOSA entities. For example, configuration like this:

```
scope: development
namePrefix: dosa.example
connector: yarpc
entityPaths:
    - ./entities
    - ./models
```
Enables users to bypass any notion of a registry or connector setup by using the new `dosaclient` package:

```
import (
    "github.com/uber-go/dosa/client"
    "github.com/uber-go/dosa/config"
)

// assume cfg is unmarshalled from YAML above...

// create and initialize client
client, err := dosaclient.New(&cfg)
```


client/client.go
```diff
@@ -0,0 +1,80 @@
...
+	case yarpc.Name():
+		conn, connErr = dosa.GetConnector(yarpc.Name(), map[string]interface{}{
+			"transport":   cfg.Yarpc.Transport,
+			"host":        cfg.Yarpc.Host,
+			"port":        cfg.Yarpc.Port,
+			"callername":  cfg.Yarpc.CallerName,
+			"servicename": cfg.Yarpc.ServiceName,
+		})
+	default:
+		return nil, fmt.Errorf("unknown connector type: %s - must be one of: devnull, memory, random or yarpc", cfg.Connector)
```
rkuris: can't we just call dosa.GetConnector(cfg.Connector) and if that fails, then render the error it returns? I think it returns something sensible when the connector is not registered.

The only thing we have to deal with is that the second argument is nil for everything except yarpc...

connectors/base/base.go
```diff
@@ -188,3 +190,8 @@ func (c *Connector) Shutdown() error {
 	}
 	return c.Next.Shutdown()
 }
+
+// Name returns the name of the connector
+func Name() string {
```
rkuris: Every connector needs this, not just those that implement base, so let's make this part of the Connector interface

client/client.go
eculver: I agree, it's a little clunky. Ideally, the `GetConnector` would take variadic arguments that mutates a configuration object. Something like this:
```
type ConnectorOption func(map[string]interface{}) error

func Host(host string) ConnectorOption {
    return func(opts map[string]interface{}) error {
        opts['host'] = host
        return nil
    }
}
```
And then `GetConnector` would look like this:

```
func GetConnector(name string, options...ConnectorOption) (Connector, error) {
    ...
}
```
And the appropriate options can then be applied when the connector is registered:

```
RegisterConnector("yarpc", Host(cfg.Host), ServiceName(cfg.ServiceName))
```
Something like that?

connectors/base/base.go
eculver: In order to call `Name` on a connector, the connector has to be created, but the name is needed to create the connector which is why it's at the package level and not part of the interface.

client/client.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/uber-go/dosa/connectors/random"
+	"github.com/uber-go/dosa/connectors/yarpc"
+	"github.com/uber-go/dosa/registry"
+)
+
+// New creates a new DOSA client from the configuration provided. Depending
+// on the configuration, this method will create the appropriate connector
+// and will try to find DOSA entities to register before returning an
+// initialized client or an error. See the config package for defaults.
+func New(cfg *config.Config) (dosa.Client, error) {
```
eculver: 👍

eculver:
@kurtis - to your question about chaining connectors, the result should still be a single composite connector type that itself does the chaining when it's registered. A "chain" of connectors would involve calling the base connector's `NewConnector` method.

connectors/base/base.go
rkuris: I don't fully understand this comment. Connectors "register" themselves when the init() function on the connector is called. Then, to create the connector, you just call dosa.GetConnector with the name. At that point, the name is already registered with the connector registry.

client/client.go
rkuris: I'm not sure I follow. You don't want to register a connector with a specific Host and ServiceName, because you might have two connectors with the same name and configured differently. This will be way more likely with the Cassandra connector to talk to two different datastores.

We should probably chat about this some. I think we have the same idea but perhaps the implementation is just a bit different.

client.go
```diff
@@ -561,6 +561,12 @@ type EntityErrors struct {
 	warns []error
 }
 
+// NewEntityErrors returns a wrapper for errors encountered while parsing
+// entity struct tags.
+func NewEntityErrors(warns []error) *EntityErrors {
+	return &EntityErrors{warns: warns}
```
eculver: In the new registry, because it's searching for entities rather than being given valid entities, I wanted to re-use the existing `dosa.EntityErrors` type but because `warns` is unexported, I need a constructor to call from this new package.

client.go
```diff
@@ -561,6 +561,12 @@ type EntityErrors struct {
 	warns []error
 }
 
+// NewEntityErrors returns a wrapper for errors encountered while parsing
+// entity struct tags.
+func NewEntityErrors(warns []error) *EntityErrors {
```
junchaowu: why "errors" plural?

client/client.go
junchaowu: for config, can you not pass pointer?  Config should just pass by value, so you have some default values.

client/client.go
```diff
@@ -0,0 +1,80 @@
...
+	reg, err := registry.NewRegistrar(cfg)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not register DOSA entities in %s", cfg.EntityPaths)
+	}
+	// create connection from config
+	var conn dosa.Connector
+	var connErr error
+	switch cfg.Connector {
+	case devnull.Name():
+		conn, connErr = dosa.GetConnector(devnull.Name(), nil)
```
junchaowu: why don't you just call dosa.GetConnector(cfg.Connector)

client/client.go
junchaowu: agree with Ron's first comment

connectors/base/base.go
junchaowu: can this be a const variable instead of method?

client.go
eculver: Because it may contain multiple errors :) but, see the [`FindEntities`](https://github.com/uber-go/dosa/blob/master/finder.go#L41) method which generates these errors.

client/client.go
```diff
@@ -45,30 +40,23 @@ func New(cfg *config.Config) (dosa.Client, error) {
...
-			"host":        cfg.Yarpc.Host,
-			"port":        cfg.Yarpc.Port,
-			"callername":  cfg.Yarpc.CallerName,
-			"servicename": cfg.Yarpc.ServiceName,
-		})
-	default:
-		return nil, fmt.Errorf("unknown connector type: %s - must be one of: devnull, memory, random or yarpc", cfg.Connector)
+
+	if cfg.Connector == nil {
+		return nil, errors.Errorf("Connector configuration is nil")
```
rkuris: nit: Errorf with no formatters

client/client.go
```diff
@@ -45,30 +40,23 @@ func New(cfg *config.Config) (dosa.Client, error) {
...
+	}
+
+	args := dosa.CreationArgs(cfg.Connector)
+	connName, ok := args["name"].(string)
+	if !ok {
+		return nil, errors.Errorf("Connector config must contain 'name' (%v)", args)
```
rkuris: bad error message; if name is not a string, ok will be false, but it will contain the name

client/client.go
```diff
@@ -45,30 +40,23 @@ func New(cfg *config.Config) (dosa.Client, error) {
...
 	}
-	if connErr != nil {
-		return nil, errors.Wrapf(err, "could not create connector for type %s", cfg.Connector)
+
+	// create connector with args
+	conn, err := dosa.GetConnector(connName, args)
+	if err != nil {
+		return nil, errors.Wrapf(err, "GetConnector failed for args: %v", args)
```
rkuris: which connector failed?

PR #164, opened 2017-06-01
Add support for UberFx

eculver:
This is the final piece of the puzzle that will enable users to use DOSA v2 as an UberFx module (depends on #162 & #163). For example:

The default entity search path is `./entities/dosa`, so assuming you have an entity in `./entities/dosa/user.go` that looks like:

```
type User struct {
    dosa.Entity `dosa:"primaryKey=UUID"
    UUID dosa.UUID
    Name string
}
```
And configuration:
```
modules:
  yarpc:
    tchannel:
      listenOn: localhost
    inbounds:
      - tchannel:
          port: 9876
    outbounds:
      - service: dosa-dev-gateway
        tchannel:
          host: 127.0.0.1
          port: 21300
storage:
  dosa:
    scope: development
    namePrefix: dosa.example
    serviceName: dosa-dev-gateway
```
You can create an UberFx service that initializes a DOSA v2 client:
```
package service

import (
    "go.uber.org/fx/service"
    dosa "github.com/uber-go/dosa/client"
)

func New() (*service.Builder, error) {
    svc := &service.Builder{}
    svc.
        WithOptions(
            service.WithDependencies(
                dosa.New,
                ...
            ),
        )
    return svc, nil
}
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,11 +23,23 @@ package config
 import (
 	"time"
 
+	"go.uber.org/fx/service"
+
+	"github.com/pkg/errors"
 	"github.com/uber-go/dosa"
 	"github.com/uber-go/dosa/connectors/yarpc"
 )
 
 const (
+	// configuration path in dotted notation, primarily used for UberFx
+	cfgPath = "storage.dosa"
+
+	// use dev gateway for any non-production scope
+	// TODO(eculver): this switching should be done on server-side
+	prodScope   = "production"
+	prodService = "dosa-gateway"
+	devService  = "dosa-dev-gateway"
+
 	_defConn = "yarpc"
 
 	// default search location and what to exclude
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -23,11 +23,23 @@ package config
 import (
 	"time"
 
+	"go.uber.org/fx/service"
```
rkuris: uberfx cannot be a dependency of dosa

fx/client_test.go
```diff
@@ -0,0 +1,117 @@
...
+		Outbounds: yarpc.Outbounds{
+			"dosa-dev-gateway": {
+				Unary: ts.NewSingleOutbound("127.0.0.1:21300"),
+			},
+		},
+	}
+	return yarpc.NewDispatcher(ycfg)
+}
+
+// TODO(eculver): these tests need some love, particularly in mocking out the
```
rkuris: Suggestion: add a config that uses the null connector and test that way. That will at least test all the client code without testing the configuration of yarpc.

glide.yaml
```diff
@@ -18,8 +18,14 @@ import:
   - gomock
 - package: github.com/yookoala/realpath
 - package: github.com/jessevdk/go-flags
+- package: go.uber.org/fx
```
rkuris: move this to a new project

registry/registry.go
```diff
@@ -49,7 +49,7 @@ func (r *registrar) NamePrefix() string {
 // Find looks at its internal index to find a registration that matches the
 // entity instance provided. Return an error when not found.
 func (r *registrar) Find(entity dosa.DomainObject) (*dosa.RegisteredEntity, error) {
-	name := reflect.TypeOf(entity).Name()
+	name := reflect.TypeOf(entity).Elem().Name()
```
rkuris: This needs a test. It could never have worked. Additionally, test the case where they have the same entity name in two different packages, which I think will fail because you are stripping the package part.

This is unlikely, but possible, since they can rename the underlying table using the `name=` attribute.

rkuris:
We are not merging this here, it's going into a new project

PR #165, opened 2017-06-01
NullString and NullFloat64

umerazad:
NullTime and NullUUID to follow.

### DIFF
--- a/NullString.go
+++ b/NullString.go
@@ -1,0 +1,115 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


junchaowu:
I may miss something here. Why don't we just support pointer and use pointer for the null case?

umerazad:
@junchaowu - Using pointers was one possible solution but I decided against it mainly because it could've resulted in brittle and error-prone code. For example, imagine two structs
`type Customer struct {
        Name string
        SSN   string
        Address *string
        Phone *string
 }`

Using above structure results in a higher cognitive overhead as the user needs to remember which fields are pointers vs not. Instead when we use proper types like NullString the chances of making a mistake are lower plus there's better editor support as well. 

We can talk more in person if you feel strongly about it.

rkuris:
@junchaow nothing stops us from adding pointers in another release. Should be a lot easier once the framework is finished

PR #166, opened 2017-06-02
add RemoveRange to yarpc connector

klnusbaum:
This is part of a broader effort to enable the removal of a range of values.  The Yarpc connector is the last connector that does not have a RemoveRange reciever function on it. Once the yarpc connector has this function, we can actually alter the Connector interface to include a RemoveRange function and then add the functionality to the client.

### DIFF
--- a/connectors/yarpc/yarpc.go
+++ b/connectors/yarpc/yarpc.go
@@ -325,7 +325,7 @@ func (c *Connector) Remove(ctx context.Context, ei *dosa.EntityInfo, keys map[st
 func (c *Connector) RemoveRange(ctx context.Context, ei *dosa.EntityInfo, columnConditions map[string][]*dosa.Condition) error {
 	conditions, err := createRpcConditions(columnConditions)
 	if err != nil {
-		errors.Wrap(err, "RemoveRange failed")
+		return errors.Wrap(err, "RemoveRange failed")
 	}
 
 	request := &dosarpc.RemoveRangeRequest{
@@ -350,7 +350,7 @@ func (c *Connector) Range(ctx context.Context, ei *dosa.EntityInfo, columnCondit
 	rpcMinimumFields := makeRPCminimumFields(minimumFields)
 	conditions, err := createRpcConditions(columnConditions)
 	if err != nil {
-		errors.Wrap(err, "Range failed")
+		return nil, "", errors.Wrap(err, "Range failed")
 	}
 	rangeRequest := dosarpc.RangeRequest{
 		Ref:          entityInfoToSchemaRef(ei),
--- a/connectors/yarpc/yarpc.go
+++ b/connectors/yarpc/yarpc.go
@@ -329,7 +329,7 @@ func (c *Connector) RemoveRange(ctx context.Context, ei *dosa.EntityInfo, column
 	}
 
 	request := &dosarpc.RemoveRangeRequest{
-		Ref: entityInfoToSchemaRef(ei),
+		Ref:        entityInfoToSchemaRef(ei),
 		Conditions: conditions,
 	}
 
--- a/connectors/yarpc/yarpc.go
+++ b/connectors/yarpc/yarpc.go
@@ -319,7 +319,24 @@ func (c *Connector) Remove(ctx context.Context, ei *dosa.EntityInfo, keys map[st
 		return errors.Wrap(err, "YARPC Remove failed")
 	}
 	return nil
+}
 
+// RemoveRange removes all entities within the range specified by the columnConditions.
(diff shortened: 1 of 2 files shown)


connectors/yarpc/yarpc.go
```diff
@@ -360,6 +365,27 @@ func (c *Connector) Range(ctx context.Context, ei *dosa.EntityInfo, columnCondit
 	return results, *response.NextToken, nil
 }
 
+func createRpcConditions(columnConditions map[string][]*dosa.Condition) ([]*dosarpc.Condition, error) {
```
rkuris: This has to be named createRPCConditions to pass lint

connectors/yarpc/yarpc.go
```diff
@@ -314,7 +314,24 @@ func (c *Connector) Remove(ctx context.Context, ei *dosa.EntityInfo, keys map[st
 		return errors.Wrap(err, "YARPC Remove failed")
 	}
 	return nil
+}
+
+// RemoveRange removes all entities within the range specified by the columnConditions.
+func (c *Connector) RemoveRange(ctx context.Context, ei *dosa.EntityInfo, columnConditions map[string][]*dosa.Condition) error {
+	conditions, err := createRpcConditions(columnConditions)
+	if err != nil {
+		errors.Wrap(err, "RemoveRange failed")
```
rkuris: missing return statement

connectors/yarpc/yarpc.go
```diff
@@ -326,27 +343,15 @@ func (c *Connector) MultiRemove(ctx context.Context, ei *dosa.EntityInfo, multiK
...
-				return nil, "", errors.Wrap(err, "Bad range value")
-			}
-			rpcConditions = append(rpcConditions, &dosarpc.Condition{
-				Op:    encodeOperator(condition.Op),
-				Field: &dosarpc.Field{Name: &fieldName, Value: &dosarpc.Value{ElemValue: rv}},
-			})
-		}
+	conditions, err := createRpcConditions(columnConditions)
+	if err != nil {
+		errors.Wrap(err, "Range failed")
```
rkuris: Missing return statement

glide.lock
```diff
@@ -1,21 +1,54 @@
 hash: 7d1e85d070980bf11cab0f4dc869c2af64ecfaf41b1ccb52eaa5db06184b23ba
-updated: 2017-05-03T14:11:01.313181441-07:00
+updated: 2017-06-02T11:28:58.980126559-07:00
```
rkuris: nit: if you're just updating glide.lock, without updating glide.yaml, put that in a separate commit, with an explanation comment as to why, or at least mention it in the check-in comments. I figured it out -- you wanted the new stuff from the idl -- but it took some snooping.

glide.lock
```diff
@@ -58,7 +95,7 @@ imports:
   - version
   - wire
 - name: go.uber.org/yarpc
-  version: 6ae533f0810337028ef055b690360e050aa13219
+  version: 6b9245d126758870690c44f27f695cc7e811f913
```
rkuris: this one worries me a little -- shouldn't we be pinning yarpc to a specific tag, instead of always grabbing master?

connectors/yarpc/yarpc_test.go
```diff
@@ -703,6 +703,43 @@ func TestConnector_Range(t *testing.T) {
...
+			Value: int64(10),
+			Op:    dosa.Eq,
+		}},
+	})
+	assert.NoError(t, err)
+
+	// perform a generic error request
+	mockedClient.EXPECT().RemoveRange(ctx, gomock.Any()).Return(errors.New("test error")).Times(1)
+	err = sut.RemoveRange(ctx, testEi, map[string][]*dosa.Condition{"c2": {&dosa.Condition{
+		Value: float64(3.3),
```
rkuris: nit: 3.3 is already a float64, you shouldn't need the cast. This isn't true for int64 of course, but it is for floats.

klnusbaum: ah, this was just copy pasta. will fix.

glide.lock
klnusbaum: you're absolutely right. I normally do that. I should've just done that in this case. Will do in the future.

glide.lock
klnusbaum: I think I'm going to pull the glide update out into a separate commit and address this issue as well.

connectors/yarpc/yarpc.go
klnusbaum: ugh. how did I miss this. This is just silly. Same with the one below. I'm going to write tests to exercise these.

glide.lock
klnusbaum: turns out yarpc is pinned to version ^1.7.1

PR #169, opened 2017-06-06
Validate caller name before invoking YARPC dispatcher

eculver:
RE: [DOSA-1011](https://jira.uberinternal.com/browse/DOSA-1011) - If `$USER` contains invalid characters and subsequently, a caller name result in `dosacli-$USER`, invoking YARPC dispatcher causes a panic. 

To reproduce:
```
USER=evan.culver.test dosa scope create eculver_test # panics
```
To fix the issue, I create two new flag types, one for caller and the other for scope. Each of these flags do the "sanitizing" of incoming values for their respective flag. For caller, '.' is replaced by hyphen (since underscores are not allowed) and for scope, '.' is replaced by underscores.

For example, the example above would create the scope "eculver_test" by calling the dosa-dev-gateway as "dosacli-evan-culver-test".

Similarly for schema operations, this:

```
USER=evan.culver.test dosa schema upsert --prefix personal.eculver ./entities
```

Would attempt to upsert schema found in `./entities` by calling dosa-dev-gateway as "dosa-cli-evan-culver-test" to a scope named "evan_culver_test".

### DIFF
--- a/cmd/dosa/scope_test.go
+++ b/cmd/dosa/scope_test.go
@@ -51,7 +51,6 @@ func TestScope_ServiceDefault(t *testing.T) {
 			os.Args = []string{
 				"dosa",
 				"--service", tc.serviceName,
-				"--caller", "foo.bar",
 				"--connector", "devnull",
 				"scope",
 				cmd,
--- a/cmd/dosa/scope_test.go
+++ b/cmd/dosa/scope_test.go
@@ -51,6 +51,7 @@ func TestScope_ServiceDefault(t *testing.T) {
 			os.Args = []string{
 				"dosa",
 				"--service", tc.serviceName,
+				"--caller", "foo.bar",
 				"--connector", "devnull",
 				"scope",
 				cmd,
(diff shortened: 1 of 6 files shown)


cmd/dosa/scope_test.go
```diff
@@ -51,6 +51,7 @@ func TestScope_ServiceDefault(t *testing.T) {
 			os.Args = []string{
 				"dosa",
 				"--service", tc.serviceName,
+				"--caller", "foo.bar",
```
klnusbaum: why does this test pass? shouldn't this caller name be invalid?

rkuris:
We probably want a better general fix for this, but it's better than what we had before.

eculver:
As @rkuris and I discussed, there is still a problem by using `$USER` in any defaults. I have created [DOSA-1060](https://jira.uberinternal.com/browse/DOSA-1060) to follow-up. The details:

This is a follow up to [DOSA-1011](https://jira.uberinternal.com/browse/DOSA-1011) and [DOSA-1009](https://jira.uberinternal.com/browse/DOSA-1009). The accepted fix for those does still leave a couple cases that are not ideal. Specifically, whenever we use the value of `$USER` in a default (eg. for caller or scope) and that value contains anything other than "dash-delimited lower-case ASCII alphanumeric letters" – then commands will fail. This means that any unicode character in `$USER`, say "müeller", the CLI will still fail.
We should find a way to either replace any of these invalid characters (not just ".") with something reasonable.

PR #174, opened 2017-06-07
Fix registry and add missing coverage

eculver:
This fixes a bug where the new registry package used for auto-discovery wasn't dereferencing correctly. Along the way, I noticed that we were using `fmt.Errorf` in a couple places so I went ahead and replaced them with `errors.Errorf`. And I added missing test coverage.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -21,7 +21,6 @@
 package registry
 
 import (
-	"fmt"
 	"reflect"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


registry/registry.go
```diff
@@ -49,7 +48,7 @@ func (r *registrar) NamePrefix() string {
 // Find looks at its internal index to find a registration that matches the
 // entity instance provided. Return an error when not found.
 func (r *registrar) Find(entity dosa.DomainObject) (*dosa.RegisteredEntity, error) {
-	name := reflect.TypeOf(entity).Name()
+	name := reflect.TypeOf(entity).Elem().Name()
```
junchaowu: which test will trigger this bug?

rkuris: +1 to comment by @junchaowu -- this needs some kind of test added to make sure it doesn't break again. It's clear this code is not getting covered.

eculver: The tests I added cover this case. If I revert this one line, the tests fail.

eculver: it's hard to see in the diff, but line 81 of `./registry/registry_test.go` covers this case.

PR #175, opened 2017-06-07
[easy] fix error message wraping

klnusbaum:
Looks like there was some copy pasta when wrapping errors for the `ScanEverything` function. Let's fix these wraps and save some developer sanity.


client.go
```diff
@@ -421,12 +421,12 @@ func (c *client) ScanEverything(ctx context.Context, sop *ScanOp) ([]DomainObjec
 	// look up the entity in the registry
 	re, err := c.registrar.Find(sop.object)
 	if err != nil {
-		return nil, "", errors.Wrap(err, "Range")
+		return nil, "", errors.Wrap(err, "ScanEverything")
```
junchaowu: can you make it, failed to scaneverything?

PR #176, opened 2017-06-10
Add nulltypes to parser

umerazad:
This change adds basic 4 nullable types to entity parser (JIRA-926/929). It includes a few bug fixes to existing code as well. This change also covers parts of the request path. Response path along with remaining Yarpc Connector changes will be covered in following CRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,6 +81,8 @@ vendor/
 # JIRA plugin
 atlassian-ide-plugin.xml
 
+# VSCode
+.vscode/
 # Crashlytics plugin (for Android Studio and IntelliJ)
 com_crashlytics_export_strings.xml
 crashlytics.properties
--- a/.gitignore
+++ b/.gitignore
@@ -81,8 +81,6 @@ vendor/
 # JIRA plugin
 atlassian-ide-plugin.xml
 
-# VSCode
-.vscode/
 # Crashlytics plugin (for Android Studio and IntelliJ)
 com_crashlytics_export_strings.xml
 crashlytics.properties
@@ -171,6 +169,13 @@ cscope.out
 cscope.in.out
 cscope.po.out
 
+
+### VisualStudioCode ###
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
 
 # End of https://www.gitignore.io/api/go,node,intellij+iml,tags
 
--- a/cmd/dosa/schema_test.go
+++ b/cmd/dosa/schema_test.go
@@ -278,6 +278,14 @@ func TestSchema_Upsert_Happy(t *testing.T) {
 	exit = func(r int) {
 		assert.Equal(t, 0, r)
 	}
+
+	nameMap := map[string]bool{
+		"awesome_test_entity":           true,
+		"testnullableentity":            true,
+		"named_import_entity":           true,
+		"testnullablenamedimportentity": true,
+	}
+
 	dosa.RegisterConnector("mock", func(dosa.CreationArgs) (dosa.Connector, error) {
 		mc := mocks.NewMockConnector(ctrl)
 		mc.EXPECT().UpsertSchema(gomock.Any(), "scope", "foo", gomock.Any()).
@@ -286,7 +294,10 @@ func TestSchema_Upsert_Happy(t *testing.T) {
 				assert.True(t, ok)
 				assert.True(t, dl.After(time.Now()))
 				assert.Equal(t, 4, len(ed))
-				assert.Equal(t, "awesome_test_entity", ed[0].Name)
+
+				for _, e := range ed {
+					assert.True(t, nameMap[e.Name])
+				}
 			}).Return(&dosa.SchemaStatus{Version: int32(1)}, nil)
 		return mc, nil
 	})
--- a/cmd/dosa/schema_test.go
+++ b/cmd/dosa/schema_test.go
@@ -35,6 +35,15 @@ import (
 	_ "github.com/uber-go/dosa/connectors/devnull"
 )
 
+func getTestEntityNameMap() map[string]bool {
+	return map[string]bool{
+		"awesome_test_entity":           true,
+		"testnullableentity":            true,
+		"named_import_entity":           true,
+		"testnullablenamedimportentity": true,
+	}
+}
+
 func TestSchema_ExpandDirectories(t *testing.T) {
 	assert := assert.New(t)
 	const tmpdir = ".testexpanddirectories"
@@ -241,7 +250,10 @@ func TestSchema_Check_Happy(t *testing.T) {
 				assert.True(t, ok)
 				assert.True(t, dl.After(time.Now()))
 				assert.Equal(t, 4, len(ed))
-				assert.Equal(t, "awesome_test_entity", ed[0].Name)
+				nameMap := getTestEntityNameMap()
+				for _, e := range ed {
+					assert.True(t, nameMap[e.Name])
+				}
 			}).Return(int32(1), nil)
 		return mc, nil
 	})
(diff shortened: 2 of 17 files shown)


finder.go
```diff
@@ -246,7 +248,7 @@ func tableFromStructType(structName string, structType *ast.StructType, packageP
 	return t, nil
 }
 
-func stringToDosaType(inType string, packagePrefix string) Type {
+func stringToDosaType(inType string) Type {
```
rkuris: Removing this means that when someone imports dosa and renames the package, it won't work correctly. Please add a test entity with dosa renamed that has the new types.

entity_parser_test.go
```diff
@@ -255,6 +302,54 @@ func TestUnsupportedType(t *testing.T) {
...
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "Nullable types are not allowed")
+}
+
+type NullFloat64PrimaryKeyType struct {
+	Entity          `dosa:"primaryKey=NullFloat64Type"`
+	NullFloat64Type NullFloat64
+}
+
+func TestNullFloat64PrimaryKeyType(t *testing.T) {
```
rkuris: Nit: these tests are all the same and could be done as one

finder_test.go
```diff
@@ -54,8 +54,8 @@ func TestNonExistentDirectory(t *testing.T) {
 
 func TestParser(t *testing.T) {
 	entities, errs, err := FindEntities([]string{"."}, []string{})
-	assert.Equal(t, 13, len(entities), fmt.Sprintf("%s", entities))
-	assert.Equal(t, 14, len(errs), fmt.Sprintf("%v", errs))
+	assert.Equal(t, 15, len(entities), fmt.Sprintf("%s", entities))
+	assert.Equal(t, 17, len(errs), fmt.Sprintf("%v", errs))
```
rkuris: I think you added 4 negative tests and one positive, so these do not add up correctly.

type.go
```diff
@@ -106,7 +118,24 @@ func FromString(s string) Type {
...
+	case TNullBool.String():
+		return TNullBool
 	default:
 		return Invalid
 	}
 }
+
+func isNullableType(t Type) bool {
+	switch t {
+	case Invalid, TNullString, TNullFloat64, TNullInt64, TNullBool:
```
rkuris: Why is Invalid nullable?

finder.go
umerazad: Good catch. Fixed.

finder_test.go
umerazad: Negative tests are related to primary keys. The positive cases are all covered by the AllTypes test.

type.go
umerazad: I've renamed the function to isValidPrimaryKeyType.

.gitignore
```diff
@@ -81,6 +81,8 @@ vendor/
 # JIRA plugin
 atlassian-ide-plugin.xml
 
+# VSCode
+.vscode/
```
rkuris: This is the generated portion of this file. Best bet is to regenerate it with [gitignore](https://www.gitignore.io/api/go,tags,intellij+iml,visualstudiocode) which adds visualstudiocode support, or alternatively add .vscode to the beginning of this file, before the autogenerated parts.

cmd/dosa/schema_test.go
```diff
@@ -284,7 +285,7 @@ func TestSchema_Upsert_Happy(t *testing.T) {
 				dl, ok := ctx.Deadline()
 				assert.True(t, ok)
 				assert.True(t, dl.After(time.Now()))
-				assert.Equal(t, 1, len(ed))
+				assert.Equal(t, 4, len(ed))
 				assert.Equal(t, "awesome_test_entity", ed[0].Name)
```
rkuris: Isn't this now a floppy test? There is nothing in the contract that says that awesome_test_entity must be first in the returned list.

doc.go
```diff
@@ -22,13 +22,13 @@
 //
 // Abstract
 //
-// :warning: DOSA is *BETA* software. It is not recommended for production use.
+// :warning: DOSA is BETA software. It is not recommended for production use.
```
rkuris: not sure why this change happened, probably differing versions of docgen?

entity.go
```diff
@@ -177,6 +177,27 @@ func (e *EntityDefinition) EnsureValid() error {
 		keyNamesSeen[c.Name] = struct{}{}
 	}
 
+	if err := e.ensureNonNullablePrimaryKeys(); err != nil {
+		return errors.Wrap(err, "Nullable types are not allowed as primary key(s)")
```
rkuris: nit: not sure I'd wrap; you're not adding any additional info to the existing error. `primary key is of nullable type: "%q"` is enough of an error

type.go
```diff
@@ -106,7 +118,24 @@ func FromString(s string) Type {
...
+	case TNullBool.String():
+		return TNullBool
 	default:
 		return Invalid
 	}
 }
+
+func isValidPrimaryKeyType(t Type) bool {
+	switch t {
+	case Invalid, TNullString, TNullFloat64, TNullInt64, TNullBool:
```
rkuris: This looks backwards. `isValidPrimaryKeyType` of `TNullFloat` returns `true` but `TNullFloat` is not a valid primary key type. Perhaps you meant `isInvalidPrimaryKeyType`?

testentity/named_import_testentity.go
```diff
@@ -0,0 +1,55 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package testentity
+
+import (
+	"time"
+
+	// Using a named import is the key change in this file. This is
+	// used to test the entity parser against named dosa imports.
```
rkuris: 👍

scan_test.go
```diff
@@ -30,16 +30,20 @@ import (
...
-	dosa.Entity `dosa:"primaryKey=BoolType"`
-	BoolType    bool
-	Int32Type   int32
-	Int64Type   int64
-	DoubleType  float64
-	StringType  string
-	BlobType    []byte
-	TimeType    time.Time
-	UUIDType    dosa.UUID
+type AllTypesScanTestEntity struct {
```
rkuris: Why was this renamed? Seems like having a duplicate name in a different package is a good test.

.gitignore
umerazad: Fixed

cmd/dosa/schema_test.go
umerazad: It wasn't floppy because the entity names were same. But I've fixed it to be more explicit and resilient.

doc.go
umerazad: Not sure. I'll look into it separately.

entity_parser_test.go
```diff
@@ -198,23 +198,27 @@ func TestMissingAnnotation(t *testing.T) {
...
+	Entity          `dosa:"primaryKey=BoolType"`
+	BoolType        bool
+	Int32Type       int32
+	Int64Type       int64
+	DoubleType      float64
+	StringType      string
+	BlobType        []byte
+	TimeType        time.Time
+	UUIDType        UUID
+	NullStringType  NullString
```
junchaowu: so we only support these four null type?

entity.go
umerazad: I've removed it.

type.go
umerazad: Ooops. Good catch. Fixed.

scan_test.go
umerazad: It took me good couple of hours to debug an issue in the walker because initially I only added the new types in AllTypes structure but apparently there were two AllTypes structures and TestParser test in (finder_test.go) would fail at random iterations. It was very confusing and painful to debug. This particular definition in scan_test.go isn't used anywhere else but it needed to be exact same structure as AllTypes defined in entity_parser_test.go.

entity_parser_test.go
umerazad: NullTime and NullGUID will be supported too.

PR #180, opened 2017-06-14
Implement the Cassandra connector

rkuris:
This connector allows for a direct connection to Cassandra from
the client side. To configure the connector, you can either
specify a yaml fragment or a gocql configuration block. The default
is to connect to 127.0.0.1:9042.

This change also adds support for testing via ccm in jenkins.

Most of the DML code is just ported over from our closed source
connector, with a few changes:
 - Removal of any remaining stapi-go code
 - Removal of any remaining UNS code
 - Changes to test infrastructure to use ccm
 - Renamed Resolver to KeyspaceMapper

Other changes in this diff:
 - Upgrade to golang 1.8
 - Increased test timeouts since creating a local C* cluster takes a while
 - Abstracted KeyspaceMapper interface for mapping
   (scope, namePrefix) -> keyspace

TODO:
 - Replication factor and strategy is hardcoded; should be calculated
   based on the cluster
 - Column type compatibility is not currently checked
 - TruncateScope is not implemented
 - Provide a better way to just specify a set of contact points
 - If a generated keyspace name is >48 characters, hash it instead
 - Scope creation can't use the keyspace mapper since it doesn't get
   the namePrefix
 - Use templates for the schema portions of the connector


connectors/cassandra/datastore_common_test.go
```diff
@@ -0,0 +1,146 @@
...
+	int64Field     = "int64field"
+	doubleField    = "doublefield"
+	blobField      = "blobfield"
+	timestampField = "timestampfield"
+	boolField      = "boolfield"
+)
+
+var testStore *Connector
+
+func initTestStore(t *testing.T) error {
```
klnusbaum: I don't think you need to return an error here, since it doesn't seem possible for this function to return an error?

connectors/cassandra/datastore_multi.go
```diff
@@ -0,0 +1,165 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
klnusbaum: I have a serious concerns with checking in commented code. It makes greping the codebase a lot harder and will like be out of dat by the time it get's uncommented. Are you planning on uncommenting this out in future revisions of this PR?

connectors/cassandra/cassandra_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/gocql/gocql"
+	"github.com/uber-go/dosa"
+	"github.com/uber-go/dosa/connectors/cassandra"
+	_ "github.com/uber-go/dosa/connectors/cassandra"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/uber-go/dosa/testentity"
+)
+
+func TestNewConnector(t *testing.T) {
```
rkuris: The difference is that assert.NoError doesn't abort the test, but I can do better than this anyway; it's inconsistent at best.

connectors/cassandra/config_test.go
```diff
@@ -0,0 +1,118 @@
...
+	expected.RetryPolicy = &gocql.SimpleRetryPolicy{NumRetries: 1}
+	socketKeepalive, _ := time.ParseDuration("100ms")
+	expected.SocketKeepalive = socketKeepalive
+	expected.MaxRoutingKeyInfo = 5000
+	expected.MaxPreparedStmts = 5000
+	expected.PageSize = 2000
+	expected.SerialConsistency = gocql.Serial
+	expected.PoolConfig = gocql.PoolConfig{HostSelectionPolicy: gocql.TokenAwareHostPolicy(gocql.RoundRobinHostPolicy())}
+	expected.HostFilter = gocql.DataCentreHostFilter("dca1")
+	assert.Equal(t, config.ClusterConfig.PoolConfig, expected.PoolConfig)
```
rkuris: I copied this code from our closed source side. I'll fix it here, but it also exists still there.

connectors/cassandra/datastore_multi.go
rkuris: This was copied from the closed source connector. At some point we will be adding multi support, but yeah, we should just delete this for now.

connectors/cassandra/embedded_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/gocql/gocql"
+	"github.com/pkg/errors"
+	"github.com/uber-go/dosa"
+	"github.com/uber-go/dosa/connectors/cassandra"
+	. "github.com/uber-go/dosa/connectors/cassandra"
+)
+
+var testConnector dosa.Connector
+
+func init() {
```
rkuris: nope, code restructuring left it behind

connectors/cassandra/embedded_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+}
+
+func GetTestConnector(t *testing.T) dosa.Connector {
```
rkuris: I'm not sure how. I want to call fatal if anything goes wrong so the test doesn't keep running.

connectors/cassandra/schema.go
```diff
@@ -37,17 +55,197 @@ func (c *Connector) TruncateScope(ctx context.Context, scope string) error {
...
+func (m *RepairableSchemaMismatchError) Error() string {
+	if m.MissingTables != nil {
+		return fmt.Sprintf("Missing %d tables (first is %q)", len(m.MissingTables), m.MissingTables[0])
+	}
+	return fmt.Sprintf("Missing %d columns (first is %q in table %q)", len(m.MissingColumns), m.MissingColumns[0].Column.Name, m.MissingColumns[0].Tablename)
+}
+
+// compareStructToSchema compares a dosa EntityDefinition to the gocql TableMetadata
+// There are two main cases, one that we can fix by adding some columns and all the other mismatches that we can't fix
+func compareStructToSchema(ed *dosa.EntityDefinition, md *gocql.TableMetadata, schemaErrors *RepairableSchemaMismatchError) error {
```
klnusbaum: You're essentially doing 3 checks in this function. First you check the partitions keys for validity. Then you check the clustering columns for validity, then you ensure all the columns exist. So what if this function was the following:
```
func compareStructToSchema(...) {
  if err = checkPartitionKeys(...); err != nil {
     return err
  } else if err = checkClusteringColumn(...); err != nil {
     return err
  } else if err = checkColumns(...); err != nil {
     return err
  }
  return nil
}
```
Then your high level algorithm reads really cleanly and the details of each check are encoded in smaller, more readable functions.

connectors/cassandra/schema.go
```diff
@@ -37,17 +55,197 @@ func (c *Connector) TruncateScope(ctx context.Context, scope string) error {
...
+			schemaErrors.MissingColumns = append(schemaErrors.MissingColumns, MissingColumn{Column: *col, Tablename: ed.Name})
+		}
+		// TODO: check column type
+	}
+
+	return nil
+
+}
+
+// CheckSchema verifies that the schema passed in the registered entities matches the database
```
klnusbaum: Maybe explain why this is only ever returning 0 or 1?

connectors/cassandra/schema.go
```diff
@@ -37,17 +55,197 @@ func (c *Connector) TruncateScope(ctx context.Context, scope string) error {
...
+	}
+	return int32(1), nil
 }
 
-// UpsertSchema is not implemented
+// UpsertSchema checks the schema and then updates it
 func (c *Connector) UpsertSchema(ctx context.Context, scope string, namePrefix string, ed []*dosa.EntityDefinition) (*dosa.SchemaStatus, error) {
-	panic("next diff")
+	sr, err := c.CheckSchema(ctx, scope, namePrefix, ed)
+	if repairs, ok := err.(*RepairableSchemaMismatchError); ok {
```
klnusbaum: This is getting a little deep. Might be worth pulling some of these things out into their own, named functions. IMHO the code would read nicely thusly:
```
if repairs, ok := err.(*RepairableSchemaMismatchError); ok {
  repairMissingTables(...)
  repairMissingColumns(...)
}
```

connectors/cassandra/schema.go
```diff
@@ -37,17 +55,197 @@ func (c *Connector) TruncateScope(ctx context.Context, scope string) error {
...
+// Error prints a human-readable error message describing the first missing table or column
+func (m *RepairableSchemaMismatchError) Error() string {
+	if m.MissingTables != nil {
+		return fmt.Sprintf("Missing %d tables (first is %q)", len(m.MissingTables), m.MissingTables[0])
+	}
+	return fmt.Sprintf("Missing %d columns (first is %q in table %q)", len(m.MissingColumns), m.MissingColumns[0].Column.Name, m.MissingColumns[0].Tablename)
+}
+
+// compareStructToSchema compares a dosa EntityDefinition to the gocql TableMetadata
+// There are two main cases, one that we can fix by adding some columns and all the other mismatches that we can't fix
```
klnusbaum: It might be worth adding to the docs here that the function only _**returns**_ an error in the case of a non-reparable mis match being found. That wasn't quite clear to me on the first pass.

connectors/cassandra/schema.go
```diff
@@ -22,12 +22,30 @@ package cassandra
...
 )
 
-// CreateScope is not implemented
+// CreateScope creates a keyspace
 func (c *Connector) CreateScope(ctx context.Context, scope string) error {
-	panic("next diff")
+	// drop the old scope, ignoring errors
+	_ = c.DropScope(ctx, scope)
+
+	ksn := CleanupKeyspaceName(scope)
```
klnusbaum: Why don't we use the keyspace mapper here to determine the keyspace?

connectors/cassandra/schema.go
```diff
@@ -37,17 +55,197 @@ func (c *Connector) TruncateScope(ctx context.Context, scope string) error {
 
 // DropScope is not implemented
 func (c *Connector) DropScope(ctx context.Context, scope string) error {
-	panic("next diff")
+	ksn := CleanupKeyspaceName(scope)
```
klnusbaum: why don't we use the keyspace mapper here to determine the keyspace?

connectors/cassandra/schema.go
```diff
@@ -37,17 +55,197 @@ func (c *Connector) TruncateScope(ctx context.Context, scope string) error {
...
+	cts := &bytes.Buffer{}
+	fmt.Fprintf(cts, `ALTER TABLE "%s"."%s" ( ADD "%s" %s )`, keyspace, table, col.Name, cassandraType(col.Type))
+	return cts.String()
+}
+
+func (c *Connector) createTable(keyspace string, ed *dosa.EntityDefinition) error {
+	return c.Session.Query(createTableString(keyspace, ed)).Exec()
+}
+
+func createTableString(keyspace string, ed *dosa.EntityDefinition) string {
```
klnusbaum: When generating CQL for all of our crud and queries we use go's templating framework, why not also use the templating here? I think it will make the code way easier to read.

connectors/cassandra/schema_test.go
```diff
@@ -0,0 +1,92 @@
...
+package cassandra
+
+import (
+	"testing"
+
+	"github.com/gocql/gocql"
+	"github.com/stretchr/testify/assert"
+	"github.com/uber-go/dosa"
+)
+
```
klnusbaum: Our code for creating a table string is way too complex to not have tests. Please add tests for that.

connectors/cassandra/embedded.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"net"
+	"strconv"
+)
+
+// CassandraPort is the default port for talking to Cassandra
+const CassandraPort = 9042
+
+// IsRunning provides a quick check to see if Cassandra is already running
+func IsRunning() bool {
```
rkuris: I purposefully made this exported so that someone could do cassandra.IsRunning() from inside their application or test code.

klnusbaum: Do we actually have a use case for this right now? If not, I feel like we shouldn't add this yet. It adds surface area to our API. We can always add it later if someone actually needs it, but why make a commitment that isn't actually needed?

connectors/cassandra/cassandra_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	sr := dosa.SchemaRef{Scope: "example", NamePrefix: "example"}
+
+	err = c.Upsert(ctx, &dosa.EntityInfo{Ref: &sr, Def: &ei.EntityDefinition}, map[string]dosa.FieldValue{
+		"an_uuid_key": dosa.UUID("c778ba9e-a241-471c-9b5b-4b4c1ef1c5b7"),
+		"strkey":      "test",
+		"int64key":    int64(11),
+	})
+
+	if err != nil {
```
rkuris: I don't think so. The test will fail when you call t.Fatal, with enough diagnostics (the error message) to tell you what the problem is, since a stack is included.

connectors/cassandra/schema.go
```diff
@@ -0,0 +1,232 @@
...
+
+	"github.com/gocql/gocql"
+	"github.com/pkg/errors"
+	"github.com/uber-go/dosa"
+)
+
+// CreateScope creates a keyspace
+func (c *Connector) CreateScope(ctx context.Context, scope string) error {
+	// drop the old scope, ignoring errors
+	_ = c.DropScope(ctx, scope)
```
klnusbaum: this seems really dangerous, to just implicitly drop a scope. Can we do better here?

(meant to leave this comment in the previous review, but it looks like it got dropped some how).

connectors/cassandra/schema.go
```diff
@@ -0,0 +1,232 @@
...
+
+	ksn := CleanupKeyspaceName(scope)
```
klnusbaum: In here, and in `DropScope`, we directly map the the scope to the keyspace name. But in `CheckSchema` we actually use the `KeySpace` mapper to determine the keyspace to use.

I'm guessing this has to do with some of the other problems related around keyspace, scope, and prefix mapping that we've talked about offline. Can we at least leave a `TODO` here to let people know that this inconsistent behavior is going to be fixed soon?

(meant to leave this comment in the previous review, but it too seemed to have gotten dropped some how).

connectors/cassandra/cassandra.go
```diff
@@ -0,0 +1,89 @@
...
+func init() {
+	dosa.RegisterConnector("cassandra", func(args dosa.CreationArgs) (dosa.Connector, error) {
+		// set up defaults for configuration
+		config := gocql.NewCluster(localhostipv4)
+		keyspaceMapper := KeyspaceMapper(DefaultKeyspaceMapper)
+
+		// parameters were provided, so lets use them
+		if args != nil {
+			// TODO the most common case is hosts-only so make that easier
+			if configYaml, ok := args["yaml"]; ok {
```
eculver: this is confusing, the config should already be yaml, so why would there be a key in the args called "yaml"? Does this mean the config would look like this:
```
connector:
  name: cassandra
  yaml: ???
```
I would suggest defining a configuration type similar to the other connectors so that it's clear what configuration is required.

connectors/cassandra/schema.go
klnusbaum: ah, you actually explained this the description for the PR. My bad.

connectors/cassandra/schema.go
rkuris: We can do better, but that's not the job of this connector to do that. There should be some additional connectors for authn and authz to prevent doing things like performing scope operations. That connector is TBD. By the time you get here, it's assumed you have appropriate rights to do this.

In the office, we should probably have a connector that holds our "rules" about not dropping production scopes. They need to be a bit more complicated though -- perhaps only dosa-group members should be able to drop production scopes. We've had a few tickets about this already.

connectors/cassandra/cassandra.go
rkuris: Ideally, we should have all the connectors take either a snippet of configuration yaml or a structure that is specific to the connector, or possibly a K/V set. The problem with the latter is that sometimes you need more depth than that. We should really clean this up though, but that's a problem for another PR.

PR #182, opened 2017-06-15
Allow entities to be provided explicitly to dosaclient.New

eculver:
This corrects an oversight in the design of the `dosaclient` interface which was the assumption that entities could be "auto-discovered" -- this is not feasible for environments where only a binary is distributed and thus there is nothing to be "discovered". This maintains backwards compatibility by allowing the associated client and registry constructors to optionally take a slice of entities that will cause entity discovery to be subverted.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -70,7 +70,16 @@ func (r *registrar) FindAll() ([]*dosa.RegisteredEntity, error) {
 }
 
 // NewRegistrar returns a new Registrar for the configuration provided.
-func NewRegistrar(cfg *config.Config) (dosa.Registrar, error) {
+func NewRegistrar(cfg *config.Config, entities ...dosa.DomainObject) (dosa.Registrar, error) {
+	// when entities are provided explicitly, discovery isn't necessary. In
+	// fact, the whole "auto-discovery" feature should only be used as a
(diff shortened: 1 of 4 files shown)


klnusbaum:
For a different PR: can we kill all autodiscovery? I can't imagine any one is actually using it since all the services here at Uber distribute binaries only.

registry/registry.go
```diff
@@ -70,7 +70,16 @@ func (r *registrar) FindAll() ([]*dosa.RegisteredEntity, error) {
...
+	// when entities are provided explicitly, discovery isn't necessary. In
+	// fact, the whole "auto-discovery" feature should only be used as a
+	// convenience for the CLI -- it's fundamentally problematic for
+	// environments where the source is not distributed alongside the binary
+	// where there is nothing to be discovered.
+	if len(entities) > 0 {
+		return dosa.NewRegistrar(cfg.Scope, cfg.NamePrefix, entities...)
+	}
+
 	idx := make(map[dosa.FQN]*dosa.RegisteredEntity)
```
klnusbaum: Can we just get rid of this code? (maybe that's for a different PR.)

eculver: Yes, that's the plan. I wanted to get the client interface fixed first though to fix the bug.

PR #183, opened 2017-06-16
index type converter

junchaowu:
Add index thrift converter 
Change indexes from array to map (it makes test easy)

### DIFF
--- a/connectors/yarpc/helpers.go
+++ b/connectors/yarpc/helpers.go
@@ -170,7 +170,7 @@ func EntityDefinitionToThrift(ed *dosa.EntityDefinition) *dosarpc.EntityDefiniti
 	}
 }
 
-// FromThriftToPrimaryKey converts thrift primary key type to doas primary key type
+// FromThriftToPrimaryKey converts thrift primary key type to dosa primary key type
 func FromThriftToPrimaryKey(key *dosarpc.PrimaryKey) *dosa.PrimaryKey {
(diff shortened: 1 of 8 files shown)


connectors/yarpc/helpers.go
```diff
@@ -136,23 +136,55 @@ func RPCTypeToClientType(t dosarpc.ElemType) dosa.Type {
...
+	}
+	return &dosarpc.EntityDefinition{
+		PrimaryKey: PrimaryKeyToThrift(ed.Key),
+		FieldDescs: fd,
+		Name:       &ed.Name,
+		Indexes:    indexes,
+	}
+}
+
+// FromThriftToPrimaryKey converts thrift primary key type to doas primary key type
```
eculver: nit: typo (s/doas/dosa)

entity_test.go
```diff
@@ -190,42 +190,34 @@ func TestEntityDefinitionEnsureValidForIndex(t *testing.T) {
 	}
 
 	invalidName := getValidEntityDefinition()
-	invalidName.Indexes[0].Name = "foo=bar"
+	invalidName.Indexes["index3=1123"] = invalidName.Indexes["index1"]
 
 	nilPK := getValidEntityDefinition()
-	nilPK.Indexes[0].Key = nil
+	nilPK.Indexes["index1"].Key = nil
```
eculver: should set the entire value to `nil`

PR #184, opened 2017-06-17
Add support for memory connector Range() tokens

rkuris:
This implements continuation tokens for the Range() operator in the
memory connector.  Just before returning from Range(), if the length
of the return value is longer than requested, we encode the data
in the last value returned to the user.  The encoding is a base64
encoded gob.

When a token is supplied on a future call to Range(), we still find
all the rows, and then decode the token, and look for that value
in the result set.  If we find it, then we just start at the next
value. If it's not there, that's okay too; we just start at the
next value in the set. The latter can only happen when the specific
value in the token has been deleted.

### DIFF
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -499,11 +499,6 @@ func (c *Connector) Shutdown() error {
 	return nil
 }
 
-// Name returns the name of the connector
-func Name() string {
-	return name
-}
-
 // NewConnector creates a new in-memory connector
 func NewConnector() *Connector {
 	c := Connector{}
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -534,15 +534,9 @@ func (c *Connector) Scan(_ context.Context, ei *dosa.EntityInfo, minimumFields [
 
 	// if there was a token, decode it so we can determine the starting
 	// partition key
-	var startPartKey string
-	var start map[string]dosa.FieldValue
-	if token != "" {
-		var err error
-		start, err = decodeToken(token)
-		if err != nil {
-			return nil, "", errors.Wrapf(err, "Invalid token %q", token)
-		}
-		startPartKey = partitionKeyBuilder(ei, start)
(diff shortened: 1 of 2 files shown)


rkuris:
Reviewers: this has been broken into three separate commits for ease of review, a cleanup commit 678ac94, the Range() work 6f9a2cf, and the Scan() work.

connectors/memory/memory_test.go
```diff
@@ -707,3 +784,93 @@ func TestPanics(t *testing.T) {
...
+	sut := NewConnector()
+
+	// we don't use the token if there's no data that matches, so lets
+	// create one row
+	createTestData(t, sut, func(id int) string {
+		return "data"
+	}, 1)
+
+	token := "this is not a token and not a hot dog"
+	t.Run("testInvalidTokenRange", func(t *testing.T) {
```
rkuris: The other test, for range on scan, is nearly identical and shares the same setup. The error report when you fail a test is really nice too.

connectors/memory/memory.go
```diff
@@ -534,15 +534,9 @@ func (c *Connector) Scan(_ context.Context, ei *dosa.EntityInfo, minimumFields [
...
-	if token != "" {
-		var err error
-		start, err = decodeToken(token)
-		if err != nil {
-			return nil, "", errors.Wrapf(err, "Invalid token %q", token)
-		}
-		startPartKey = partitionKeyBuilder(ei, start)
+	startPartKey, start, err := getStartingPoint(ei, token)
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "Invalid token %s", token)
```
klnusbaum: Nit, I think this wrap error message is incorrect. The `getStartingPoint` function could've failed for any number of reasons. I would make this wrap error message something like `failed to scan` and then (if you want) wrap the error from `decodeToken` down in `getStartingPoint`. In _that_ wrap, I would mention about the invalid token.

rkuris: Yes, that's true. I will fix it then merge.

PR #185, opened 2017-06-19
Add helper methods for index

junchaowu:


doc.go
```diff
@@ -22,13 +22,13 @@
 //
 // Abstract
 //
-// :warning: DOSA is *BETA* software. It is not recommended for production use.
+// :warning: DOSA is BETA software. It is not recommended for production use.
```
rkuris: Why is this changing?

PR #186, opened 2017-06-19
[easy] Update RangeOp.Limit docstring

eculver:
Fixes [DOSA-1049](https://jira.uberinternal.com/browse/DOSA-1049) - Docstring for RangeOp.Limit misrepresents default behavior.

### DIFF
--- a/range.go
+++ b/range.go
@@ -111,8 +111,8 @@ func (r *RangeOp) Fields(fieldsToRead []string) *RangeOp {
 	return r
 }
 
-// Limit sets the number of rows returned per call. If not set, a default
-// value would be applied
+// Limit sets the number of rows returned per call. A limit must be provided
+// otherwise an error will be returned.
 func (r *RangeOp) Limit(n int) *RangeOp {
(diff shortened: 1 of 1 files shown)


range.go
```diff
@@ -111,8 +111,8 @@ func (r *RangeOp) Fields(fieldsToRead []string) *RangeOp {
 	return r
 }
 
-// Limit sets the number of rows returned per call. If not set, a default
-// value would be applied
+// Limit sets the number of rows returned per call. A limit must be provided
+// otherwise an error will be returned.
```
klnusbaum: This function doesn't return an error. Would it make more sense to say "A limit must be provided otherwise an error will be returned by the Range function."

junchaowu: How about set a high value here, maxint for example, the server will narrow it down to 200 anyway. A value here just to prevent the error.

eculver: @klnusbaum I agree, I will update.
@junchaowu we decided to make this required to prevent unexpected behavior.

PR #187, opened 2017-06-21
Fixing doc.

umerazad:


junchaowu:
why do you add this star. Every build I have will revert this change back somehow.

PR #188, opened 2017-06-21
glide cleanup

klnusbaum:
Our glide.yaml file needed some love. I've implemented the following changes:

- the go_flags package now tracks a semantic version
- several packages have been moved to testImport since they're not actually needed to compile any of our real code
- our yarpc dependencies have been cleaned up as well. We actually specify which subpackages we need, change the semver to ^1.7.1 and get rid of the reference to github.com/uber-go/yarpc which isn't actually used anywhere in our code.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,32 +1,31 @@
 package: github.com/uber-go/dosa
 import:
+- package: github.com/elodina/go-avro
+- package: github.com/golang/mock
+  subpackages:
+  - gomock
+- package: github.com/jessevdk/go-flags
+  version: ^1.2.0
 - package: github.com/pkg/errors
   version: ^0.8.0
 - package: github.com/satori/go.uuid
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -1,32 +1,31 @@
...
+  - transport/http
+  - transport/tchannel
+testImport:
+- package: github.com/yookoala/realpath
 - package: github.com/stretchr/testify
   version: ^1.1.4
   subpackages:
   - assert
-  - mock
-- package: github.com/golang/mock
```
eculver: Shouldn't this be kept in `testImport`?

klnusbaum: I believe it's actually needed up here since we supply our own mocks. when I ran through glide's recommendations regarding our `glide.yaml` file, it said this belongs up here.

PR #190, opened 2017-06-22
add remove range to client

klnusbaum:
This adds the ability for the dosa client to call remove range

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -329,6 +329,42 @@ func TestClient_Upsert_Errors(t *testing.T) {
 	assert.Contains(t, err.Error(), "badcol")
 }
 
+func TestClient_RemoveRange(t *testing.T) {
+	reg1, _ := dosaRenamed.NewRegistrar(scope, namePrefix, cte1)
+
+	c1 := dosaRenamed.NewClient(reg1, nullConnector)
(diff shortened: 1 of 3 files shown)


client_test.go
```diff
@@ -329,6 +329,42 @@ func TestClient_Upsert_Errors(t *testing.T) {
...
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "ClientTestEntity1")
+	assert.Contains(t, err.Error(), "borkborkbork")
+
+	// success case
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	mockConn := mocks.NewMockConnector(ctrl)
+	mockConn.EXPECT().CheckSchema(ctx, gomock.Any(), gomock.Any(), gomock.Any()).Return(int32(1), nil).AnyTimes()
+	mockConn.EXPECT().RemoveRange(ctx, gomock.Any(), gomock.Any()).Return(nil)
```
eculver: you might want to use the expected range op instead of `gomock.Any()`

klnusbaum: connectors don't take a RangeOp. They take an entity definition and a map of column conditions.

PR #191, opened 2017-06-22
add index to the backward compatibility check

junchaowu:


entity.go
```diff
@@ -396,14 +395,31 @@ func (e *EntityDefinition) IsCompatible(e2 *EntityDefinition) error {
...
+			return errors.Errorf("the column %s in old entity %s but not in new entity", name, e2.Name)
 		}
 		if colType1 != colType2 {
 			return errors.Errorf("the type for column %s mismatch: (%v vs %v)", name, colType1, colType2)
 		}
 	}
 
-	// TODO Handle tags and index comparison in the future
+	// Index can only be added, not mutated
+	if len(e2.Indexes) > len(e1.Indexes) {
```
klnusbaum: do you really need this check? Won't the check on line 413 catch this and give a better error message?

junchaowu: this is the case that e2.Indexes is not nil but e1.Indexes is nil. Otherwise the 412 will crash

klnusbaum: ah ok. Can we just change the if statement to say that then? That way it will be clearer what we're checking.

PR #197, opened 2017-07-12
Memory connector secondary index support

rkuris:
Adds support for secondary indexes in memory; see individual commits for changes

### DIFF
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -267,7 +267,7 @@ func (c *Connector) Read(_ context.Context, ei *dosa.EntityInfo, values map[stri
 		return nil, &dosa.ErrNotFound{}
 	}
 
-	if len(ei.Def.ClusteringKeySet()) == 0 {
+	if len(ei.Def.Key.ClusteringKeySet()) == 0 {
 		return partitionRef[0], nil
 	}
 	// clustering key, search for the value in the set
@@ -309,7 +309,7 @@ func (c *Connector) mergedInsert(ei *dosa.EntityInfo,
 		return nil
 	}
 
-	if len(ei.Def.ClusteringKeySet()) == 0 {
+	if len(ei.Def.Key.ClusteringKeySet()) == 0 {
 		// no clustering key, so the row must already exist, merge it
 		return mergeFunc(partitionRef[0], values)
 	}
@@ -347,7 +347,7 @@ func (c *Connector) Remove(_ context.Context, ei *dosa.EntityInfo, values map[st
 	}
 
 	// no clustering keys? Simple, delete this
-	if len(ei.Def.ClusteringKeySet()) == 0 {
+	if len(ei.Def.Key.ClusteringKeySet()) == 0 {
 		// NOT delete(entityRef, encodedPartitionKey)
 		// Unfortunately, Scan relies on the fact that these are not completely deleted
 		entityRef[encodedPartitionKey] = nil
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -38,8 +38,6 @@ import (
 	"github.com/uber-go/dosa/connectors/base"
 )
 
-const name = "memory"
-
 // Connector is an in-memory connector.
 // The in-memory connector stores its data like this:
 // map[string]map[string][]map[string]dosa.FieldValue
@@ -67,12 +65,12 @@ type partitionRange struct {
 	end          int
 }
 
-// delete deletes the values referenced by the partitionRange. Since this function modifies
+// remove deletes the values referenced by the partitionRange. Since this function modifies
 // the data stored in the in-memory connector, a write lock must be held when calling
 // this function.
 //
 // Note this function can't be called more than once. Calling it more than once will cause a panic.
-func (pr *partitionRange) delete() {
+func (pr *partitionRange) remove() {
 	partitionRef := pr.entityRef[pr.partitionKey]
 	pr.entityRef[pr.partitionKey] = append(partitionRef[:pr.start], partitionRef[pr.end+1:]...)
 	pr.entityRef = nil
@@ -89,13 +87,17 @@ func (pr *partitionRange) values() []map[string]dosa.FieldValue {
 // partitionKeyBuilder extracts the partition key components from the map and encodes them,
 // generating a unique string. It uses the encoding/gob method to make a byte array as the
 // key, and returns this as a string
-func partitionKeyBuilder(pk *dosa.PrimaryKey, values map[string]dosa.FieldValue) string {
+func partitionKeyBuilder(pk *dosa.PrimaryKey, values map[string]dosa.FieldValue) (string, error) {
 	encodedKey := bytes.Buffer{}
 	encoder := gob.NewEncoder(&encodedKey)
 	for _, k := range pk.PartitionKeys {
-		_ = encoder.Encode(values[k])
+		if v, ok := values[k]; ok {
+			_ = encoder.Encode(v)
+		} else {
+			return "", errors.Errorf("Missing value for partition key %q", k)
+		}
 	}
-	return string(encodedKey.Bytes())
+	return string(encodedKey.Bytes()), nil
 }
 
 // findInsertionPoint locates the place within a partition where the data belongs.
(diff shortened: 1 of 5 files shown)


connectors/memory/memory.go
```diff
@@ -38,8 +38,6 @@ import (
 	"github.com/uber-go/dosa/connectors/base"
 )
 
-const name = "memory"
```
klnusbaum: why are we getting rid of this?

connectors/memory/memory.go
```diff
@@ -245,9 +247,20 @@ func compareType(d1 dosa.FieldValue, d2 dosa.FieldValue) int8 {
...
+	err := c.mergedInsert(ei.Def.Name, ei.Def.Key, values, func(into map[string]dosa.FieldValue, from map[string]dosa.FieldValue) error {
 		return &dosa.ErrAlreadyExists{}
 	})
+	if err != nil {
+		return err
+	}
+	for iName, iDef := range ei.Def.Indexes {
+		// this error must be ignored, so we skip indexes when the value
+		// for one of the index fields is not specified
+		_ = c.mergedInsert(iName, ei.Def.UniqueKey(iDef.Key), values, overwriteValuesFunc)
```
klnusbaum: This makes me nervous. Right now, this relies on the fact that `mergedInsert` happens to only return an error when the value for one of the index fields isn't specified. In the future, if `mergedInsert` ever returns an error for another reason, we'll being ignoring that error here. It's also not super obvious that `mergedInsert` has this behavior, as evidenced by the fact that you had to document this strange ignoring of the error.

In other words, it feels to me like you're relying on an implementation detail of `mergeInsert` rather than just on it's contract.

I think the better way to do this is just not called merged insert if the values we're given don't have the index value in them. Doing this check
  - shouldn't be very hard to write
  - shouldn't need any new tests, since the behavior is the same
  - will make the code cleaner
  - and because the check will probably have some obvious name (like `hasRequiredIndexValue`) we can get rid of this docstring because the code will be self-documenting.


If you do go this route, the for loop will become a little more complex, but also be identical to the for-loop used down in upsert. You could probably extract the code into it's on function, `updateIndicies`.

connectors/memory/memory.go
```diff
@@ -278,28 +294,40 @@ func (c *Connector) Read(_ context.Context, ei *dosa.EntityInfo, values map[stri
...
-		}
-		return nil
-	})
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if err := c.mergedInsert(ei.Def.Name, ei.Def.Key, values, overwriteValuesFunc); err != nil {
+		return err
+	}
+	for iName, iDef := range ei.Def.Indexes {
+		_ = c.mergedInsert(iName, ei.Def.UniqueKey(iDef.Key), values, overwriteValuesFunc)
```
klnusbaum: this also relies on the weird implementation details of `mergedInsert` but you forgot the doc here, making it unclear why we ignore the error. I think this is further evidence that we shouldn't call merged insert if we know it's going to fail. We should "document the behavior with code"

connectors/memory/memory.go
```diff
@@ -330,45 +358,63 @@ func (c *Connector) mergedInsert(name string,
...
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	partitionRange := c.findRange(ei, columnConditions)
+	partitionRange, err := c.findRange(ei, columnConditions)
+	if err != nil {
+		return err
+	}
+	// TODO: this should have been from a primary key lookup, not an index lookup
 	if partitionRange != nil {
```
klnusbaum: Nit: make this if statement an early return instead:
```
if partitionRange == nil {
    return nil
}
// rest of function
```

connectors/memory/memory_test.go
```diff
@@ -727,7 +752,29 @@ func TestConnector_ScanWithToken(t *testing.T) {
 		err = sut.Remove(context.TODO(), clusteredEi, data[0])
 		assert.NoError(t, err)
 	}
+}
 
+func TestConnector_ScanWithTokenFromWrongTable(t *testing.T) {
```
klnusbaum: What is this doing in this PR?

connectors/memory/memory_test.go
```diff
@@ -873,4 +935,4 @@ func createTestData(t *testing.T, sut *Connector, keyGenFunc func(int) string, i
 			"c7": dosa.FieldValue(dosa.UUID(uuid.NewV1().String()))})
 		assert.NoError(t, err)
 	}
-}
+}
```
klnusbaum: NO NEWLINE? THIS IS A TRAVESTY!!! 😠 😠 😠

Just kidding :p

connectors/memory/memory.go
```diff
@@ -245,9 +247,20 @@ func compareType(d1 dosa.FieldValue, d2 dosa.FieldValue) int8 {
 // Otherwise, search the partition for the exact same clustering keys. If there, fail
 // if not, then insert it at the right spot (sort.Search does most of the heavy lifting here)
 func (c *Connector) CreateIfNotExists(_ context.Context, ei *dosa.EntityInfo, values map[string]dosa.FieldValue) error {
-	return c.mergedInsert(ei.Def.Name, ei.Def.Key, values, func(into map[string]dosa.FieldValue, from map[string]dosa.FieldValue) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	err := c.mergedInsert(ei.Def.Name, ei.Def.Key, values, func(into map[string]dosa.FieldValue, from map[string]dosa.FieldValue) error {
 		return &dosa.ErrAlreadyExists{}
 	})
+	if err != nil {
```
klnusbaum: can this if statement be merged with line 252?
```
alreadyExistsFunc := func(into map[string]dosa.FieldValue, from map[string]dosa.FieldValue) error {
    return &dosa.ErrAlreadyExists{}
}
if err := c.mergedInsert(ei.Def.Name, ei.Def.Key, values, alreadyExistsFunc); err != nil {
  return err
}
```

you could probably even pull out the function definition (I'm not sure if the go-compiler is smart enough to realize it doesn't need to recreate that function everytime `CreateIfNotExists` is called)

range.go
```diff
@@ -197,3 +197,28 @@ func (m rangeOpMatcher) String() string {
 		m.eqScanOp.String(),
 	)
 }
+
+// IndexFromConditions returns the name of the index or the base table to use, along with the key info
+// for that index. If no suitable index could be found, an error is returned
+func (ei *EntityInfo) IndexFromConditions(conditions map[string][]*Condition) (name string, key *PrimaryKey, err error) {
```
klnusbaum: nit: This function name seems a little misleading to me, since it doesn't necessarily return an Index in the conventional sense. Not sure what would be better though... :/

range.go
```diff
@@ -197,3 +197,28 @@ func (m rangeOpMatcher) String() string {
...
+	var indexDef *IndexDefinition
+	for name, indexDef = range ei.Def.Indexes {
+		key = indexDef.Key
+		// we check the range conditions before adding the uniqueness columns
+		if err := EnsureValidRangeConditions(ei.Def, key, conditions, identityFunc); err == nil {
+			return name, ei.Def.UniqueKey(key), nil
+		}
+	}
+	// none of the indexes work, so fail
+	return "", nil, errors.Wrapf(baseTableError, "No index matches specified conditions")
```
klnusbaum: "No _primary key_ or index matches the specified condition"

connectors/memory/memory.go
rkuris: Consistency. None of the other connectors have this.

connectors/memory/memory.go
```diff
@@ -278,28 +294,40 @@ func (c *Connector) Read(_ context.Context, ei *dosa.EntityInfo, values map[stri
 	return partitionRef[inx], nil
 }
 
+func overwriteValuesFunc(into map[string]dosa.FieldValue, from map[string]dosa.FieldValue) error {
```
rkuris: It doesn't just copy, it overwrites.

connectors/memory/memory.go
klnusbaum: I've thought about this a little more, and now I'm not so sure we should change it. If you have some time today, let's chat about it offline?

rkuris: A good compromise would be to have a specific error returned from mergedInsert, and change the code to be more like:

```
// merge the values, but ignore the error when keys are missing
if err = c.mergedInsert(...); err != nil && !isMissingKeyValueError(err) {
    return err
}
```

range.go
rkuris: A more interesting question is, what is an index in this context? One could argue that the primaryKey is just another index, especially since it shares the same PrimaryKey structure.

range.go
rkuris: Same comment as above. Perhaps we can still do better here -- "A search cannot be performed with the specified criteria on this entity due to a lack of a matching index or primary key constraint" but that seems super wordy.

klnusbaum:
Chatted with @rkuris offline. The main problem I had with this diff was the issue with `mergedInsert`. We're just going to fix that later. It's being tracked in https://jira.uberinternal.com/browse/DOSA-1104

PR #200, opened 2017-07-21
Ensure Registered Entities are created with Index information

klnusbaum:
`NewRegisteredEntity` has a bug where the entity that get's registered doesn't properly copy over the secondary Indexes for a given `EntityDefintion`, it only sets the name, key, and columns.

This causes Range lookups on secondary indexes to fail for the in-memory connector.

Let's fix this problem by just using the full `EntityDefinition` that's available to us via the `table` that's passed in. This way, if we ever add any more fields to the `EntityDefinition` in the future, they'll be automatically included in the `RegisteredEntity`.

### DIFF
--- a/connectors/memory/memory_test.go
+++ b/connectors/memory/memory_test.go
@@ -540,6 +540,24 @@ func TestConnector_Range(t *testing.T) {
 	}, dosa.All(), "", 200)
 	assert.NoError(t, err)
 	assert.Empty(t, data)
+
+	// Test Ranging on an Index
+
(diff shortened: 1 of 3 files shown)


rkuris:
This description does not describe the symptoms of this bug. Can you add that as the title? What is the effect of not having the index in the registry?

Can we add a test?

klnusbaum:
Updated title and description. Added tests to the in-memory connector to ensure `Range` on indexes works.

PR #201, opened 2017-07-21
Provide convenience function for fetching entire range

klnusbaum:
Customers often want to fetch an entire range of values. Doing so with the current `Range` function provided by the `Client` is tedious. It requires writing the same boiler plate code over, and over again. Let's provide a convenience function so that customers don't have to write their own.

### DIFF
--- a/client.go
+++ b/client.go
@@ -163,9 +163,12 @@ type Client interface {
 	// of the range can be fetched with additional calls to the range function.
 	Range(ctx context.Context, rangeOp *RangeOp) ([]DomainObject, string, error)
 
-	// FetchRange fetches the entire range of values specified by the RangeOp. For each value fetched,
-	// the provided onNext function is called with the value as it's argument.
-	FetchRange(ctx context.Context, r *RangeOp, onNext func(value DomainObject) error) error
+	// RangeIter starts at the offset specified by the RangeOp and fetches the entire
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -437,6 +445,27 @@ func (c *client) Range(ctx context.Context, r *RangeOp) ([]DomainObject, string,
 	return objectArray, token, nil
 }
 
+func (c *client) FetchRange(ctx context.Context, r *RangeOp, onNext func(value DomainObject) error) error {
```
rkuris: If the inner func wants to make another database call, it will need the context passed to it

klnusbaum: If I'm calling `FetchRange` with a context, won't there be a pretty good chance the closure I'm passing in already has access to that context?

rkuris: Well, that's not perfect, but acceptable. This forces someone to embed the function inline. There are workarounds, they can just wrap the call externally if they really need it. We should probably document this behavior.

client.go
```diff
@@ -437,6 +445,27 @@ func (c *client) Range(ctx context.Context, r *RangeOp) ([]DomainObject, string,
 	return objectArray, token, nil
 }
```
rkuris: Add some docs here -- specifically around how to get the context into the function for go noobs

rkuris:
Still needs docs somewhere, or perhaps an example, but that's different from this PR.

PR #210, opened 2017-08-11
Add RoutingConnector interface

DavidCorn:
- create configuration struct and interfaces
- create basic routing connector interface without implementing methods
- error handlings are not fully done

### DIFF
--- a/connectors/routingconnector/config.go
+++ b/connectors/routingconnector/config.go
@@ -1,0 +1,33 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package routingconnector
+
+// Config is a struct contains fields from yaml
+// scope should be an exact string in any case,
+// namePrefix could be in 3 different format:
+// 1. exact string like "service" that matches namePrefix that is exactly "service"
+// 2. partial glob match like "service.*" that matches all namePrefix that has a prefix of "service."
+// 3. full glob match like "*" that matches all namePrefix
+type Config struct {
+	// the outer key for ConnectorMap is scope, inner key is namePrefix,
+	// and value is the connectorName
+	Routers []map[string]map[string]string `yaml:"routers"`
+}
--- a/connectors/routingconnector/routing_config.go
+++ b/connectors/routingconnector/routing_config.go
@@ -1,0 +1,76 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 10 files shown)


connectors/routingconnector/configured_connector.go
```diff
@@ -0,0 +1,97 @@
...
+
+// FetchConnector fetches the connector based on the name in the dosa client config
+func FetchConnector(cfg *ConnectorConfig) (dosa.Connector, error) {
+	if cfg.TargetConnector == nil || cfg.TargetConnector.Config == nil {
+		return nil, errors.Errorf("There's no dosa client configuration exists")
+	}
+
+	cConfig := cfg.TargetConnector.Config
+	if cConfig.Connector == nil {
+		return nil, errors.Errorf("Connector configuration is nil")
```
rkuris: I think this is impossible due to line 76, testing will prove it

connectors/routingconnector/config.go
```diff
@@ -0,0 +1,28 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package routingconnector
+
+// Config is a struct contains fields from yaml
+type Config struct {
+	// the outer key for ConnectorMap is scope, inner key is namePrefix,
+	// and value is the connectorName
+	Routers []map[string]map[string]string `yaml:"routers"`
```
fuj: Did you add a default config somewhere?

connectors/routingconnector/routing_config.go
```diff
@@ -0,0 +1,70 @@
...
+// RoutingConfig implements routing method
+// to decide which connector the RoutingConnector talks to
+type RoutingConfig struct {
+	Scope      string
+	NamePrefix string
+}
+
+// NewRoutingConfig initializes RoutingConfig
+func NewRoutingConfig(namePrefix string, scope string) (*RoutingConfig, error) {
+	if namePrefix == "" {
```
fuj: What about empty scope?

connectors/routingconnector/routing_config.go
```diff
@@ -0,0 +1,70 @@
...
+
+// RoutingConfig implements routing method
+// to decide which connector the RoutingConnector talks to
+type RoutingConfig struct {
+	Scope      string
+	NamePrefix string
+}
+
+// NewRoutingConfig initializes RoutingConfig
+func NewRoutingConfig(namePrefix string, scope string) (*RoutingConfig, error) {
```
fuj: nit: use the same variable order from struct definition for easier auditing

connectors/routingconnector/config.go
DavidCorn: config would be in passed in with NewRoutingConnector, and is configured and loaded in dosa-gateway

connectors/routingconnector/routing_config.go
```diff
@@ -0,0 +1,70 @@
...
+		// we don't support name like '*.service.v1'
+		return nil, errors.Errorf(
+			"namePrefix like %v is not supported, cannot put * at the beginning of namePrefix.", namePrefix)
+	}
+
+	return &RoutingConfig{NamePrefix: namePrefix, Scope: scope}, nil
+}
+
+// RoutesTo implements the method to choose the matched connector
+func (rc *RoutingConfig) RoutesTo(scope string, namePrefix string) bool {
```
fuj: Again, make the order of `scope` and `namePrefix` consistent

connectors/routingconnector/routing_config.go
```diff
@@ -0,0 +1,70 @@
...
+	// scope should be a exact match
```
fuj: `an exact`

connectors/routingconnector/routing_config.go
```diff
@@ -0,0 +1,70 @@
...
+	if rc.Scope != scope {
+		return false
+	}
+
+	// namePrefix could be glob match
```
fuj: the line below is an exact match. move this comment to line #64 before actual glob match code?

connectors/routingconnector/routing_connector.go
```diff
@@ -0,0 +1,304 @@
...
+	"context"
+
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/uber-go/dosa"
+	"github.com/uber-go/dosa/connectors/base"
+)
+
+// PluginFunc is a plugin function that takes scope, namePrefix and operation name, then gives wanted scope and namePrefix
```
fuj: multi-line comment?

connectors/routingconnector/routing_config.go
DavidCorn: Empty scope shouldn't matter because it will be an exact match, and won't affect the sorting logic, and I think it should be there since config yaml file is static. but yeah I can add it here for safe

eculver:
I strongly suggest we move this to a different package namespace. The current namespace violates all of the Golang idioms. "connectors/routingconnector/routing_connector.go" stutters twice. If anyone uses the `RoutingConnector` type, it's three time. This is nutz:

```
import "github.com/uber-go/dosa/connectors/routingconnector" // stutter

func MyFunction() {
    routingconnector.RoutingConnector{
        // no
    }

    routingconnector.NewRoutingConfig {
        // y so many "routing"?
    }
}
```

Please please please can we change the package name to "routing" and just use types like `Config` and `Connector` in that package. This is what it would look like:

```
import "github.com/uber-go/dosa/connectors/routing" // 0-stutter

func MyFunction() {
    routing.Connector{
        // no stutter
    }

    routing.NewConfig{
        // so much better
    }
}
```

fuj:
+1 to what @eculver said above

PR #212, opened 2017-08-16
add version into rpc header

junchaowu:


version.go
```diff
@@ -0,0 +1,24 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package dosa
+
+// VERSION indicates the dosa client version
+const VERSION = "2.0"
```
rkuris: This is not the dosa client version, it's the yarpc version only. I'm not sure we want to use this for other connectors or use cases.

connectors/yarpc/yarpc_test.go
```diff
@@ -181,7 +182,7 @@ func TestYaRPCClient_Read(t *testing.T) {
 	}
 
 	// we expect a single call to Read, and we return back two fields, f1 which is in the typemap and another field that is not
-	mockedClient.EXPECT().Read(ctx, readRequest).Return(&drpc.ReadResponse{drpc.FieldValueMap{
+	mockedClient.EXPECT().Read(ctx, readRequest, gomock.Any()).Return(&drpc.ReadResponse{drpc.FieldValueMap{
```
rkuris: Use something more specific than gomock.Any() to verify that the correct version is being passed

junchaowu: Sorry I can't. because they calloption is an interface, golang can't compare two interface.

version.go
junchaowu: I think everything is relied on DOSA version. It gives us the extensibility that other connectors can also benefits from it.

PR #217, opened 2017-08-23
Normalize yarpc connector args

eculver:
This is a minor user-experience enhancement. Because we're passing in a raw map of key/value pairs as the arguments to the method that creates connectors, the keys are case-sensitive which is not very intuitive. This changes the behavior so that users can provide raw configuration yaml like:
```
connector:
    name: yarpc
    host: 127.0.0.1
    port: "1234"
    serviceNAmE: "dosa-gateway"
```
Without having to worry about whether the keys should be camel-case or not.

In doing so, I also added coverage for the yarpc connector's `RegisterConnector` method.

### DIFF
--- a/connectors/yarpc/yarpc.go
+++ b/connectors/yarpc/yarpc.go
@@ -630,19 +630,25 @@ func getWithDefault(args map[string]interface{}, elem string, def string) string
 
 func init() {
 	dosa.RegisterConnector("yarpc", func(args dosa.CreationArgs) (dosa.Connector, error) {
-		host, ok := args["host"]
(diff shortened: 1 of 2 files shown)


klnusbaum:
Can we try, instead, just providing a more descriptive error message. e.g "Here are the keys that are required. Here are the ones you gave us".

eculver:
Per feedback from @kurtis we have decided to opt out of this in favor of better documentation for the YARPC connector. The biggest issue is that the behavior is specific to YARPC instead of being generic for all connectors.

PR #222, opened 2017-09-01
update router structure

junchaowu:


connectors/routing/connector.go
```diff
@@ -44,46 +42,21 @@ type RoutingConnector struct {
...
 // NewRoutingConnector initializes the RoutingConnector
 // connectorMap has a key of connectorName, and the value is a dosa.connector instance
-func NewRoutingConnector(cfg Config, connectorMap map[string]dosa.Connector, plugin PluginFunc) (*RoutingConnector, error) {
-	cConnectors, err := NewConnectorConfigs(cfg, connectorMap)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not initialize ConfiguredConnectors")
-	}
-	connSlice := sortConnectorSlice(cConnectors)
-
+func NewRoutingConnector(cfg Config, connectorMap map[string]dosa.Connector, plugin PluginFunc) *RoutingConnector {
```
DavidCorn: since it's in routing/connectors.go, shall we just change it to NewConnector()?

connectors/routing/config.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return nil, fmt.Errorf("can't find any router info for scope %s, namePrefix %s", scope, namePrefix)
+}
+
+
+// FindDefaultRouter finds the default router information.
+func (c *Config) FindDefaultRouter() (*RoutingConfig, error) {
+	for _, router := range c.Routers {
+		if router.Scope == DefaultScope {
```
DavidCorn: change the const DefaultScope = "default" into this file?

connectors/routing/config.go
```diff
@@ -0,0 +1,110 @@
...
+				router, err := NewRoutingConfig(scope, namePrefixStr, connectorName)
+				if err != nil {
+					return errors.Wrap(err, "failed to parse routing config")
+				}
+				routers = append(routers, router)
+			}
+		}
+	}
+	sort.Sort(routers)
+	*r = routers
```
DavidCorn: If r is initialized inside UnmarshalYAML, why not just return like (*Router, error)?

junchaowu: UnmarshalYAML is a predefined interface that yaml library will use to load yaml config. I can't change the interface.

connectors/routing/config.go
```diff
@@ -0,0 +1,110 @@
...
+
+
+// FindDefaultRouter finds the default router information.
+func (c *Config) FindDefaultRouter() (*RoutingConfig, error) {
+	for _, router := range c.Routers {
+		if router.Scope == DefaultScope {
+			return router, nil
+		}
+	}
+	return nil, errors.New("there should be a default scope defined in routing config yaml file")
```
rkuris: Can we catch this error earlier, like when we load the configuration?

connectors/routing/connector.go
rkuris: +1

connectors/routing/connector.go
```diff
@@ -104,23 +77,24 @@ func (rc *RoutingConnector) getConnector(scope string, namePrefix string, opName
...
-	return rc._getDefaultConnector()
-}
 
-func (rc *RoutingConnector) _getDefaultConnector() (dosa.Connector, error) {
-	for _, cConfig := range rc.CConfigs {
-		if cConfig.Config.Scope == DefaultScope {
-			return cConfig.Connector, nil
-		}
+	if err != nil {
+		return nil, errors.Wrap(err, "can't find connector")
```
rkuris: This error message is not informative enough. Which connector could not be found?

connectors/routing/connector.go
```diff
@@ -104,23 +77,24 @@ func (rc *RoutingConnector) getConnector(scope string, namePrefix string, opName
...
 	}
-	return nil, errors.New("there should be a default scope defined in routing config yaml file")
+
+	c, ok := rc.connectors[router.Connector]
+	if !ok {
+		return nil, errors.Wrapf(err, "can't find %s connector", router.Connector)
```
rkuris: nit: use %q here so the string will be quoted (in case of leading or trailing spaces)

connectors/routing/config.go
```diff
@@ -0,0 +1,110 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package routing
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
```
eculver: nit: goimports

connectors/routing/config.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
eculver: It looks like this go added twice

connectors/routing/config.go
```diff
@@ -0,0 +1,142 @@
...
+}
+func (r Routers) Less(i, j int) bool {
+	if r[i].Scope == r[j].Scope {
+		return r[i].NamePrefix > r[j].NamePrefix
+	}
+
+	return r[i].Scope > r[j].Scope
+}
+
+// UnmarshalYAML unmarshals the config into gocql cluster config
```
eculver: Suggest adding an example configuration that this handles

connectors/routing/config_test.go
```diff
@@ -0,0 +1,116 @@
...
+# - [scope]
+#    [namePrefix_1]: connectorName
+#    [namePrefix_2]: connectorName
+- production:
+    default: cassandra
+    serviceA: cassandra
+- development:
+    default: cassandra
+    serviceB: cassandra
+- eats:
```
eculver: Suggest not using real service names.

connectors/routing/connector_test.go
```diff
@@ -252,24 +222,22 @@ func TestGetConnector(t *testing.T) {
 
 func TestRoutingConnector_CreateIfNotExists(t *testing.T) {
```
eculver: nit: should replace "Routing" here to be consistent

PR #225, opened 2017-09-21
Redis connector

tiffanycheng:
v0 of redis connector implementation

### DIFF
--- a/connectors/redis/redis.go
+++ b/connectors/redis/redis.go
@@ -1,0 +1,240 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package redis
+
+import (
+	"context"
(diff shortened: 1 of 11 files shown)


connectors/redis/redis.go
```diff
@@ -0,0 +1,210 @@
...
+
+// ErrNotImplemented is returned when a user didn't call Initialize
+type ErrNotImplemented struct{}
+
+// Error returns a constant string "client not initialized"
+func (*ErrNotImplemented) Error() string {
+	return "Not implemented"
+}
+
+// ErrInvalidEntity is returned when a user didn't call Initialize
```
rkuris: incorrect comment

testutil/portcheck.go
```diff
@@ -0,0 +1,14 @@
+package testutil
```
rkuris: please move this portcheck change into it's own commit

rkuris:
A problem with this diff is that it uses closed source code, so we will either need to close source this connector or use a public domain variant.

connectors/redis/redis.go
```diff
@@ -0,0 +1,255 @@
...
+// Connector for redis database
+type Connector struct {
+	base.Connector
+	client SimpleRedis
+	ttl    time.Duration
+}
+
+// CreateIfNotExists not implemented
+func (c *Connector) CreateIfNotExists(ctx context.Context, ei *dosa.EntityInfo, values map[string]dosa.FieldValue) error {
+	return new(ErrNotImplemented)
```
rkuris: Although not needed, this is pretty easy to implement, except that it requires a change to the SimpleRedis interface

connectors/redis/redis.go
```diff
@@ -0,0 +1,255 @@
...
+	return nil, "", new(ErrNotImplemented)
+}
+
+// Scan not implemented.
+func (c *Connector) Scan(ctx context.Context, ei *dosa.EntityInfo, minimumFields []string, token string, limit int) (multiValues []map[string]dosa.FieldValue, nextToken string, err error) {
+	return nil, "", new(ErrNotImplemented)
+}
+
+// CheckSchema not implemented.
+func (c *Connector) CheckSchema(ctx context.Context, scope string, namePrefix string, ed []*dosa.EntityDefinition) (version int32, err error) {
```
rkuris: remove any of the schema operations; this will cause the requests to go to the base connector, which returns a different error (NoMoreConnectors). That's more appropriate for these schema errors.

connectors/redis/redis_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+func TestWriteReadKeyValue(t *testing.T) {
+	if !redis.IsRunning() {
+		t.Skip("Redis is not running")
+	}
+
+	table, _ := dosa.TableFromInstance(&testentity.KeyValue{})
+	sr := dosa.SchemaRef{Scope: "example", NamePrefix: "example"}
+	testEi := &dosa.EntityInfo{Ref: &sr, Def: &table.EntityDefinition}
+	values := map[string]dosa.FieldValue{
```
rkuris: nit: these keep repeating so perhaps they should be declared outside of all the tests

connectors/routing/config.go
```diff
@@ -100,7 +100,7 @@ type Config struct {
 }
 
 // FindRouter finds the router information based on scope and namePrefix.
-func (c *Config) FindRouter(scope, namePrefix string) (*Rule) {
+func (c *Config) FindRouter(scope, namePrefix string) *Rule {
```
rkuris: this belongs in a different commit if you intend to change this (probably not a bad idea)

PR #227, opened 2017-09-26
Deprecate Search API

rkuris:
We actually don't need Search or SearchOps, so lets delete them.
Range() works on indexes already.


rkuris:
@klnusbaum good catch, I regenerated the mocks using the correct version of mockgen

PR #229, opened 2017-09-29
Schema reducing connector

tiffanycheng:
Have a bunch of questions that I left as TODO comments


connectors/schemaadapter/schema_reducer.go
```diff
@@ -0,0 +1,171 @@
...
+	type rangeResults struct {
+		token string
+		rows []map[string]dosa.FieldValue
+	}
+
+	unpack := rangeResults{}
+	err = json.Unmarshal(response, &unpack)
+	return unpack.rows, unpack.token, err
+
+	// TODO who is supposed to be calling the redis connector Upsert with the range response adapted?
```
tiffanycheng: It seems like the fallback cache connector has to take care of most of this adaptation logic, not this schema reducing connector. 

For example, for this range method, the fallback cache connector takes the range response, creates a key for it, and either calls the redis connector's upsert or the redis connector's read. I no longer see how this SchemaReducing connector fits into the flow.

PR #231, opened 2017-10-05
Fallback cache connector

tiffanycheng:


connectors/cache/fallback.go
```diff
@@ -0,0 +1,222 @@
...
+		Columns: []*dosa.ColumnDefinition{
+			{Name: value, Type: dosa.Blob},
+			{Name: key, Type: dosa.Blob},
+		},
+	}
+	return adaptedEi
+}
+
+// used for single entry reads/writes
+func createCacheKey(ei *dosa.EntityInfo, values map[string]dosa.FieldValue, e Encoder) []byte {
```
tiffanycheng: @eculver my plan is to order the keys alphabetically so that i can just use gob encoding instead of thrift. Thrift is pretty complicated as I would have to adapt everything over into the thrift objects manually

tiffanycheng:
i am thinking it might be a good idea to have a kill switch that says don't use the fallback at all. i could either build it into this connector, or we could have yet another connector that sits on top of this one that basically switches between this one and the original connector

connectors/cache/encoding.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// NewJSONEncoder returns a json encoder
+func NewJSONEncoder() Encoder {
+	return &jsonEncoder{}
+}
+
+type jsonEncoder struct{}
+
+func (j *jsonEncoder) Encode(v interface{}) ([]byte, error) {
```
eculver: Missing some linting here re: doc strings

connectors/cache/fallback.go
```diff
@@ -0,0 +1,272 @@
...
+	encoder           Encoder
+	cacheableEntities map[string]bool
+	// Used primarily for testing so that nothing is called in a goroutine
+	synchronous bool
+}
+
+// SetCachedEntities sets the entities that will write and read from the fallback
+func (c *Connector) SetCachedEntities(entities ...dosa.DomainObject) {
+	set := createCachedEntitiesSet(entities)
+	c.cacheableEntities = set
```
eculver: is it ok that this isn't thread-safe? this should only be called once on initialization or no?

connectors/cache/fallback.go
```diff
@@ -0,0 +1,306 @@
...
+		cacheValue, err := c.encoder.Encode(values)
+		if err != nil {
+			return err
+		}
+		adaptedEi := adaptToKeyValue(ei)
+		newValues := map[string]dosa.FieldValue{
+			key:   cacheKey,
+			value: cacheValue,
+		}
+		return c.fallback.Upsert(ctx, adaptedEi, newValues)
```
eculver: I think calls to the fallback should extend the parent context to prevent spurious timeouts caused by high latency in the origin's response.

connectors/cache/fallback.go
```diff
@@ -0,0 +1,306 @@
...
+		t, err := dosa.TableFromInstance(e)
+		if err != nil {
+			continue
+		}
+		set[t.EntityDefinition.Name] = true
+	}
+	return set
+}
+
+func adaptToKeyValue(ei *dosa.EntityInfo) *dosa.EntityInfo {
```
eculver: suggest adding a comment here (or changing the name of the method) to describe that a new entity info is being derived from the original to support the structure of a caching connector

PR #232, opened 2017-10-10
Fix RemoveRange panic with index criteria

rkuris:
It's illegal to remove multiple items based on an index. This change
enforces that, by not looking at indexes when calling RemoveRange.

### DIFF
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -403,11 +403,10 @@ func (c *Connector) RemoveRange(_ context.Context, ei *dosa.EntityInfo, columnCo
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	partitionRange, err := c.findRange(ei, columnConditions)
+	partitionRange, err := c.findRange(ei, columnConditions, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


connectors/memory/memory.go
```diff
@@ -485,7 +484,7 @@ func decodeToken(token string) (values map[string]dosa.FieldValue, err error) {
 //
 // Note that this function reads from the connector's data map. Any calling functions should hold
 // at least a read lock on the map.
-func (c *Connector) findRange(ei *dosa.EntityInfo, columnConditions map[string][]*dosa.Condition) (*partitionRange, error) {
+func (c *Connector) findRange(ei *dosa.EntityInfo, columnConditions map[string][]*dosa.Condition, searchIndexes bool) (*partitionRange, error) {
```
klnusbaum: boolean flag arguments make me sad :( http://www.informit.com/articles/article.aspx?p=1392524

klnusbaum:
Here's another link, this one actually explains different ways to get rid of boolean parameters: https://mortoray.com/2015/06/15/get-rid-of-those-boolean-function-parameters/

PR #234, opened 2017-10-19
Add hit rate stats to redis connector

tiffanycheng:


connectors/redis/redis.go
```diff
@@ -40,6 +41,19 @@ type SimpleRedis interface {
 	Shutdown() error
 }
 
+// Scope is a namespace wrapper around a stats reporter, ensuring that
+// all emitted values have a given prefix or set of tags.
+type Scope interface {
+	// Counter returns the Counter object corresponding to the name.
+	Counter(name string) tally.Counter
```
klnusbaum: IMHO would shouldn't bring in the `tally` dependency here. It ties people to a very specific stats logging framework. This line is the only reason we need to import it, and so I'd say the `Counter` function shouldn't return a `tally.Counter`, but instead an interface that we define.

This will of course mean that we can't directly pass a tally counter to the constructor for this connector (bummer). We'll have to write a little, thin adapter so that a Tally scope works with a "dosa" scope.

connectors/redis/redis.go
```diff
@@ -40,6 +41,19 @@ type SimpleRedis interface {
 	Shutdown() error
 }
 
+// Scope is a namespace wrapper around a stats reporter, ensuring that
+// all emitted values have a given prefix or set of tags.
+type Scope interface {
```
klnusbaum: `Scope` isn't something that's tied to the redis connector specifically. We may have other connectors where we want to do logging as well. I would say let's put this interface in it's own package.

connectors/redis/redis.go
```diff
@@ -212,7 +220,34 @@ func (c *Connector) Remove(ctx context.Context, ei *dosa.EntityInfo, keys map[st
...
+	if err != nil {
+		c.incStat("error", method)
+	}
+}
+
+func (c *Connector) incStat(subscope, method string) {
+	if c.stats == nil {
+		return
+	}
+	c.stats.SubScope("cache").SubScope(subscope).Tagged(map[string]string{"method": method}).Counter("redis").Inc(1)
```
klnusbaum: `"redis"` seems like a weird name for the counter. I would actually use the variable `subscope` in it's position (and remove it's earlier usage). I'd also not call it subscope. Maybe the following:
```
func (c *Connector) incStat(action, method string) {
 	if c.stats == nil {
 		return
 	}
 	c.stats.SubScope("cache").Tagged(map[string]string{"method": method}).Counter(action).Inc(1)
}
```

connectors/redis/redis_test.go
```diff
@@ -164,9 +161,6 @@ func TestReadNotFound(t *testing.T) {
 		t.Skip("Redis is not running")
 	}
 
-	table, _ := dosa.TableFromInstance(&testentity.KeyValue{})
```
klnusbaum: In the future please pull refactoring changes like this out into their own PR. it makes things easier to review.

metrics/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+	Counter(name string) Counter
+
+	// Tagged returns a new child scope with the given tags and current tags.
+	Tagged(tags map[string]string) Scope
+
+	// SubScope returns a new child scope appending a further name prefix.
+	SubScope(name string) Scope
+
+	// Timer returns the Timer object corresponding to the name.
+	Timer(name string) Timer
```
klnusbaum: do we need Time for anything yet? If we're not going to have a use for it in any upcomming PRs, I'd say let's hold off adding it for now.

connectors/redis/redis_test.go
```diff
@@ -232,9 +215,134 @@ func TestEntityTypeNotBytes(t *testing.T) {
...
+	var rc = redis.NewConnector(testRedisConfig, nil)
 	err := rc.Shutdown()
 	assert.NoError(t, err)
 	err = rc.Shutdown()
 	assert.NoError(t, err)
 }
+
+// Test that a successful read from redis logs as cache hit
+func TestReadHitStat(t *testing.T) {
+	if !redis.IsRunning() {
```
klnusbaum: This tests seem to have some significant copy pasta. Consider refactoring, perhaps pulling out the common stuff into a single function. You also _might_ be able to write a single test that has a for loop in it that just iterates over each of the different permutations you want to test. Like: https://github.com/uber-go/dosa/blob/master/connectors/cassandra/schema_test.go#L95

PR #235, opened 2017-10-24
Latency stats for redis commands

tiffanycheng:


connectors/redis/redigo.go
```diff
@@ -88,6 +90,11 @@ func (c *simpleRedis) Shutdown() error {
 // Do is a proxy method that calls Redigo's `Do` method and returns its output. It remembers
 // to close connections taken from the pool
 func (c *simpleRedis) do(commandName string, args ...interface{}) (interface{}, error) {
+	if c.stats != nil {
```
klnusbaum: Let's write a `NoOp` impelmentation of the stats interface so we don't have to do all these nil checks.

connectors/redis/redigo_test.go
```diff
@@ -73,8 +75,39 @@ func TestWrongPort(t *testing.T) {
...
+
+	c := redis.NewRedigoClient(testRedisConfig.ServerSettings, stats)
+
+	redisCommands := map[string]func(){
+		"GET": func() { c.Get("a") },
+		"SET": func() { c.SetEx("a", []byte{1}, 9*time.Second) },
+		"DEL": func() { c.Del("a") },
+	}
+	for command, f := range redisCommands {
+		stats.EXPECT().SubScope("redis").Return(stats)
```
klnusbaum: nit: run each of these as a seperate test using `t.Run`. E.g. https://github.com/uber-go/dosa/blob/master/connectors/cassandra/schema_test.go#L142

connectors/redis/redigo.go
tiffanycheng: i'll get this in first and then add a noop separately and change all the nil checks at once

PR #236, opened 2017-10-25
Update dependencies, fix breaking change in gocql

Willyham:


connectors/yarpc/yarpc_test.go
```diff
@@ -812,12 +812,12 @@ func TestConnector_Scan(t *testing.T) {
 			},
 			NextToken: &responseToken,
 		}, nil)
-	// failed call, return error
```
klnusbaum: Why do these need to be changed?

klnusbaum:
Overall I really like this change. I just wanna understand a little better why we need to change the yarpc test. That seems orthogonal to all this.

klnusbaum:
This is a good fix. Please update the `changelog` file as well. We'll put this in a patch version. so please create a 2.3.1 entry in the change log file.

eculver:
my vote is for fixing gocql/gocql (by reverting ab2b96a) -- this becomes a no-op and would benefit the greater gocql community

eculver:
This: https://github.com/gocql/gocql/pull/1002

klnusbaum:
I don't think we need this any more thanks to @eculver 's fix. We're actually probably going to move the cassandra connector to it's own repository in a bit in order to mitigate this issue all together. So I don't think we need this.

PR #239, opened 2017-10-30
Redis connector nil ptr check

DavidCorn:
Return NoOp default scope when stats is nil


metrics/metrics.go
```diff
@@ -33,3 +53,51 @@ type Timer interface {
 	// Stop reports time elapsed since the timer start to the recorder.
 	Stop()
 }
+
+// CheckStats checks whether stats is nil. If it is nil then return default scope
+func CheckIfNilStats(stats Scope) Scope {
```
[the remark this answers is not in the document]
DavidCorn: 🆗

connectors/redis/redigo.go
```diff
@@ -90,11 +90,10 @@ func (c *simpleRedis) Shutdown() error {
 // Do is a proxy method that calls Redigo's `Do` method and returns its output. It remembers
 // to close connections taken from the pool
 func (c *simpleRedis) do(commandName string, args ...interface{}) (interface{}, error) {
-	if c.stats != nil {
-		t := c.stats.SubScope("redis").SubScope("latency").Timer(commandName)
-		t.Start()
-		defer t.Stop()
-	}
+	t := metrics.CheckIfNilStats(c.stats).SubScope("redis").SubScope("latency").Timer(commandName)
```
tiffanycheng: call checkifnilstats in NewRedigoClient

metrics/no_op_stats.go
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
tiffanycheng: tests?

metrics/no_op_stats.go
```diff
@@ -0,0 +1,75 @@
...
+// NoOpScope scopes nothing
+type NoOpScope struct {}
+
+// NoOpCounter Counts nothing
+type NoOpCounter struct {}
+
+// NoOpTimer times nothing
+type NoOpTimer struct {}
+
+func (s *NoOpScope) Counter(name string) Counter {
```
eculver: lint: missing docstring.

metrics/no_op_stats.go
```diff
@@ -0,0 +1,75 @@
...
+// CheckStats checks whether stats is nil. If it is nil then return default scope
+func CheckIfNilStats(stats Scope) Scope {
+	if stats == nil {
+		return &NoOpScope{}
+	}
+	return stats
+}
+
+// NoOpScope scopes nothing
+type NoOpScope struct {}
```
eculver: nit: this shouldn't be camel-cased

metrics/no_op_stats.go
```diff
@@ -0,0 +1,75 @@
...
+func (s *NoOpScope) SubScope(name string) Scope {
+	return &NoOpScope{}
+}
+
+func (s *NoOpScope) Timer(name string) Timer {
+	return &NoOpTimer{}
+}
+
+func (c *NoOpCounter) Inc(delta int64) {
+	// do nothing
```
eculver: nit: move these to the docstring: `// Inc is a noop`

metrics/no_op_stats.go
eculver: suggest calling this just `noop.go` for simplicity.

metrics/no_op_stats.go
```diff
@@ -0,0 +1,75 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package metrics
+
+import (
+	"time"
+)
+
```
eculver: run `gofmt` -- extra lines will go away.

connectors/redis/redis.go
```diff
@@ -96,7 +96,7 @@ func NewConnector(config Config, scope metrics.Scope) dosa.Connector {
 	return &Connector{
 		client: NewRedigoClient(config.ServerSettings, scope),
 		ttl:    config.TTL,
-		stats:  scope,
+		stats:  metrics.CheckIfNilStats(scope),
```
eculver: When would it make sense to pass a `nil` value here? Why not make`scope` required and if its value is `nil`, panic?

DavidCorn: We want the scope operation to work as normal even it is nil, by adding the CheckIfNilStats, we don't need to check status every time we want to use the `stats`.

metrics/no_op_stats.go
DavidCorn: Yeah will add test if the logic looks fine

PR #240, opened 2017-10-31
add testclient to DOSA

klnusbaum:
This creates a simple DOSA client useful for testing that can't be done with simple mocks. This client utilizes the in-memory connector as it's backing store.

### DIFF
--- a/testclient/testclient.go
+++ b/testclient/testclient.go
@@ -1,0 +1,16 @@
+package testclient
+
+import (
+	"github.com/uber-go/dosa"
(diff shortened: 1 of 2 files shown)


klnusbaum:
In my experience, many engineers don't know how to take that code and apply it to their project. It's also not very reusable code. This just removes all ambiguity and I can just say in our support room now "here, call this function, it gives you a test client, all setup". So I was going to do this, and then absolutely update the docs to refer to this instead :)

PR #244, opened 2017-11-10
Support time.Time in the in-memory connector

ztstewart:
This diff registers time.Time in the `makeToken` and `decodeToken` functions, allowing the `Scan` and other functions to support `time.Time`.

Fixes T1305491

### DIFF
--- a/connectors/memory/memory_test.go
+++ b/connectors/memory/memory_test.go
@@ -86,6 +86,27 @@ var clusteredEi = &dosa.EntityInfo{
 	},
 }
 
+var clusteredByTimeEi = &dosa.EntityInfo{
+	Ref: &testSchemaRef,
+	Def: &dosa.EntityDefinition{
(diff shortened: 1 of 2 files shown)


connectors/memory/memory_test.go
```diff
@@ -86,6 +86,27 @@ var clusteredEi = &dosa.EntityInfo{
...
+		},
+		Key: &dosa.PrimaryKey{
+			PartitionKeys: []string{"f1"},
+			ClusteringKeys: []*dosa.ClusteringKey{
+				{Name: "c1", Descending: false},
+				{Name: "c2", Descending: true},
+			},
+		},
+		Name: "t3",
+		Indexes: map[string]*dosa.IndexDefinition{
```
klnusbaum: Nit: Probably don't need an index for this `EntityInfo` since we never test index functionality with this entity.

PR #246, opened 2017-11-16
Ensure index name is exported in the entity during schema validation 

jaricftw:
When upserting the schema, an unexported index filed will be silently ignored. Adding this validation to error loud to the client.

### DIFF
--- a/finder_test.go
+++ b/finder_test.go
@@ -53,10 +53,10 @@ func TestNonExistentDirectory(t *testing.T) {
 }
 
 func TestParser(t *testing.T) {
-	entities, errs, err := FindEntities([]string{"."}, []string{})
+	entities, _, err := FindEntities([]string{"."}, []string{})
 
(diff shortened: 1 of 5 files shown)


finder_test.go
```diff
@@ -54,57 +54,43 @@ func TestNonExistentDirectory(t *testing.T) {
 
 func TestParser(t *testing.T) {
 	entities, errs, err := FindEntities([]string{"."}, []string{})
+	expectedEntities := map[string]DomainObject{
```
klnusbaum: Nit: This change looks unrelated to the issue in the PR. Is that true? If so, in the future, can we put unrelated changes in their own PR? It just makes things easier to review.

jaricftw: Yeah, sure. 
I added it when debugging the test failure after the change in `ParseIndexTag`, was confused by the hardcoded number.  

It's implicitly related though, i.e. the `SingleIndexUnExported` I added above won't be part of `expectedEntities` but will instead increase the errs count from 20 to 21

PR #247, opened 2017-11-28
Update the CLI Admin client to use CanUpsertSchema RPC for checking schema compatibility

jaricftw:
This pull request is based on https://github.com/uber/dosa-idl/pull/22.

- rename the `CheckSchema` method in admin client to `CanUpsertSchema`
- add a separate `CanUpsertSchema` method in the `Connector` interaface
- pick up the corresponding `CanUpsertSchema` RPC for checking schema compatibility through CLI


connector.go
```diff
@@ -125,8 +125,13 @@ type Connector interface {
...
 	// DDL operations (schema)
 	// CheckSchema validates that the set of entities you have provided is valid and registered already
-	// It returns a list of SchemaRef objects for use with later DML operations.
+	// It returns the latest schema version for use with later DML operations.
 	CheckSchema(ctx context.Context, scope string, namePrefix string, ed []*EntityDefinition) (version int32, err error)
+	// CheckSchemaToUpsert is used by the CLI to validate whether new entities
+	// can be upserted by checking their compatibility with the latest applied schema.
+	// If compatible, the latest schema version is returned,
+	// which can be "invalid" (i.e. -1) if no schema has been applied yet.
+	CheckSchemaToUpsert(ctx context.Context, scope string, namePrefix string, ed []*EntityDefinition) (version int32, err error)
```
klnusbaum: Nit: `ed` should be renamed to `eds` since it's a slice of entity definitions.

connectors/yarpc/yarpc.go
```diff
@@ -478,29 +478,41 @@ func (c *Connector) Scan(ctx context.Context, ei *dosa.EntityInfo, minimumFields
...
+	response, err := c.checkSchema(ctx, scope, namePrefix, eds, true)
 	if err != nil {
-		return dosa.InvalidVersion, wrapError(err, "CheckSchema failed", scope, c.Config.ServiceName)
+		return dosa.InvalidVersion, wrapError(err, "CheckSchema compatibility failed", scope, c.Config.ServiceName)
 	}
 
 	return *response.Version, nil
 }
 
+func (c *Connector) checkSchema(ctx context.Context, scope, namePrefix string, eds []*dosa.EntityDefinition, toUpsert bool) (*dosarpc.CheckSchemaResponse, error) {
```
klnusbaum: See my comments in https://github.com/uber/dosa-idl/pull/22. If we make `CheckSchema` and `CheckSchemaToUpsert` two different calls, then this function probably shouldn't exist.

glide.lock
```diff
@@ -1,5 +1,5 @@
-hash: 44c1b4016b4c802a043b2955cb24e2922ae9949f2ebcbbb0351529107d9d2f11
-updated: 2017-11-16T16:14:50.837044899-08:00
+hash: 2ff54d6684eff3a6f08b8735dade82b31597521ada7ef5a87002a8c5dd13acfd
```
klnusbaum: Let's keep glide updates as separate commits.

mocks/client.go
```diff
@@ -1,23 +1,3 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
```
klnusbaum: Copyright notices shouldn't be deleted. Check how you generated this.

client.go
```diff
@@ -551,7 +551,7 @@ func (c *adminClient) CheckSchema(ctx context.Context, namePrefix string) (*Sche
 	if err != nil {
 		return nil, errors.Wrapf(err, "GetSchema failed")
 	}
-	version, err := c.connector.CheckSchema(ctx, c.scope, namePrefix, defs)
+	version, err := c.connector.CheckSchemaToUpsert(ctx, c.scope, namePrefix, defs)
```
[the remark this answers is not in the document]
jaricftw: Replaced `CheckSchema` with `CanUpsertSchema`

connector.go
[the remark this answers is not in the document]
jaricftw: Yeah, I couldn't find a better name, changed to `CanUpsertSchema`

jaricftw: done

glide.lock
jaricftw: This is needed to pick up the idl update.

connectors/yarpc/yarpc.go
jaricftw: right. Also I've changed the https://github.com/uber/dosa-idl/pull/22 to use two rpc calls

connector.go
```diff
@@ -125,8 +125,13 @@ type Connector interface {
 
 	// DDL operations (schema)
 	// CheckSchema validates that the set of entities you have provided is valid and registered already
-	// It returns a list of SchemaRef objects for use with later DML operations.
-	CheckSchema(ctx context.Context, scope string, namePrefix string, ed []*EntityDefinition) (version int32, err error)
+	// It returns the latest schema version for use with later DML operations.
+	CheckSchema(ctx context.Context, scope string, namePrefix string, eds []*EntityDefinition) (version int32, err error)
+	// CanUpsertSchema is used by the CLI to validate whether new entities
```
klnusbaum: nit: don't mention the CLI here. users of the function may change over time.

glide.yaml
```diff
@@ -12,13 +12,14 @@ import:
...
 - package: github.com/uber/dosa-idl
+  version: CheckSchemaRequest-toUpsert
   subpackages:
   - .gen/dosa
   - .gen/dosa/dosaclient
 - package: go.uber.org/yarpc
   version: ^1.7.1
   subpackages:
-  - api/transport
+  - api/EntityDefsToThrift
```
klnusbaum: What's going on here. This seems incorrect.

connector.go
jaricftw: removed.

glide.yaml
jaricftw: No idea why it appeared there.. Fixed it.

cmd/dosa/schema.go
```diff
@@ -128,7 +127,7 @@ type SchemaCheck struct {
 
 // Execute executes a schema check command
```
klnusbaum: Let's update the documentation for this and the `SchemaCheck` struct. In fact, we may want to call it something different now since it's checking if the schema can be upgraded.

jaricftw: Yeah, I will do the doc change.

PR #250, opened 2017-11-30
Add context value to CheckSchema RPC to differentiate requesters, i.e DOSA CLI v.s. DOSA client

jaricftw:
CheckSchema RPC is used in two different cases:

1. in DOSA CLI, to validate whether a new entity can be upserted (i.e. the new entity is compatible with the latest schema)
2. during client initialization, to verify whether the client provide entity are valid and have been already registered. (i.e. the latest schema is compatible with the provided entity)

Here we simply pass the requester information in the context so that DOSA gateway can behave accordingly.

### DIFF
--- a/client.go
+++ b/client.go
@@ -102,6 +102,18 @@ func ErrorIsAlreadyExists(err error) bool {
 	return ok
 }
 
+type contextKey string
+
+const (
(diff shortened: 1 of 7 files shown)


jaricftw:
Closing this PR as we've decided to go with the other approach in https://github.com/uber-go/dosa/pull/247

PR #252, opened 2017-11-30
Adding a TestMain in Cassandra connector package to setup and drop scope once

jaricftw:
We are seeing many test failures with the same error of complaining not able to add an already-existing keyspace. e.g.

>  Unable to create keyspace "example": Cannot add existing keyspace "example"

This is because in each unit test, we drop the scope and recreate the scope and schema without synchronization, which could run into the above race condition. 

Adding a TestMain to setup the cassandra, create the test scope and schema before tests, and drop the scope after tests finish.

### DIFF
--- a/connectors/cassandra/schema.go
+++ b/connectors/cassandra/schema.go
@@ -36,7 +36,7 @@ func (c *Connector) CreateScope(ctx context.Context, scope string) error {
 	// drop the old scope, ignoring errors
 	err := c.DropScope(ctx, scope)
 	if err != nil {
-		fmt.Errorf("drop scope failed: %v", err)
+		fmt.Printf("drop scope error: %v, ignore it if it does not exist", err)
 	}
 	ksn := CleanupKeyspaceName(scope)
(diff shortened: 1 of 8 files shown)


connectors/cassandra/schema.go
```diff
@@ -34,11 +34,13 @@ import (
 // CreateScope creates a keyspace
 func (c *Connector) CreateScope(ctx context.Context, scope string) error {
 	// drop the old scope, ignoring errors
-	_ = c.DropScope(ctx, scope)
-
+	err := c.DropScope(ctx, scope)
+	if err != nil {
+		fmt.Errorf("drop scope failed: %v", err)
```
rkuris: This isn't really an error, so downgrade this to at least a warning, or perhaps just info.

jaricftw: yeah, I was trying to just warn it. Cannot find how we do warning here since there is no logger configured in this repo. 

`Info` is fine too.

jaricftw: Changed to just print out the error info.

PR #253, opened 2017-12-01
Log both success and error counts on redis connector functions

tiffanycheng:


junchaowu:
why both

tiffanycheng:
turns out it's hard to get a success rate when i'm only keeping track of errors

connectors/redis/redis.go
```diff
@@ -221,25 +221,23 @@ func (c *Connector) Remove(ctx context.Context, ei *dosa.EntityInfo, keys map[st
...
 func (c *Connector) logHitRate(method string, err error) {
-	if err != nil {
-		if dosa.ErrorIsNotFound(err) {
-			c.incStat("miss", method)
-			return
-		}
-		c.logError(method, err)
+	if err != nil && dosa.ErrorIsNotFound(err) {
+		c.incStat("miss", method)
 		return
```
jaricftw: This `return` should be removed? Otherwise we don't log the error here.

tiffanycheng: So for redis, I am trying to distinguish the case where you try to read a key but the key is not in redis from the case where you try to do a read but redis has some sort of i/o error. I guess we could increment both, but then it looks like we're double counting # of queries

PR #254, opened 2017-12-01
Tag fallback stats with dosa entity name

tiffanycheng:


tiffanycheng:
It’s helpful to know which cached entities are read from the fallback. Otherwise, we only know that we are reading from the fallback but not for which objects

PR #257, opened 2017-12-08
Add fallback pattern flag to router plugin

DavidCorn:


junchaowu:
where is the requirement coming from?

PR #258, opened 2017-12-13
detect whether error is thrift error or network error in yarpc connector

junchaowu:


connectors/yarpc/yarpc.go
```diff
@@ -323,7 +339,11 @@ func (c *Connector) MultiRead(ctx context.Context, ei *dosa.EntityInfo, keys []m
 
 	response, err := c.Client.MultiRead(ctx, request, VersionHeader())
 	if err != nil {
-		return nil, errors.Wrap(err, "MultiRead failed")
+		if !dosarpc.Dosa_MultiRead_Helper.IsException(err) {
+			return nil, errors.Wrap(err, "failed to multiread due to network issue")
```
phliar: Here "multiread" is not an english word; so it should either be MultiRead (which I believe is what the endpoint is called) or "multi-read". My preference is that it should be the endpoint name, so "MultiRead".

Similarly for the other errors, it would be better to put in the method name exactly, preserving case...

junchaowu: updated

connectors/yarpc/yarpc.go
```diff
@@ -532,7 +577,7 @@ func (c *Connector) UpsertSchema(ctx context.Context, scope, namePrefix string,
 	}
 
 	if response.Version == nil {
-		return nil, errors.New("UpsertSchema failed: server returns version nil")
+		return nil, errors.New("failed to upsertschema: server returns version nil")
```
phliar: Missed one! :)

PR #266, opened 2018-01-05
Change fallback connector to use only gob encoding

tiffanycheng:
The fallback connector had a bug where it was not returning values as pointers. yarpc does this, so if the fallback is supposed to be a drop in replacement, it needs to have the same behavior

### DIFF
--- a/connectors/cache/fallback_test.go
+++ b/connectors/cache/fallback_test.go
@@ -57,16 +57,6 @@ var (
 	}
 )
 
-type BadEncoder struct{}
-
-func (b *BadEncoder) Encode(interface{}) ([]byte, error) {
-	return nil, errors.New("Encoding failed")
-}
-
-func (b *BadEncoder) Decode([]byte, interface{}) error {
-	return errors.New("Decoding failed")
-}
-
 func createTestEi(sr dosa.SchemaRef) *dosa.EntityInfo {
 	table, _ := dosa.TableFromInstance(&testentity.TestEntity{})
 	testEi := &dosa.EntityInfo{Ref: &sr, Def: &table.EntityDefinition}
@@ -74,7 +64,6 @@ func createTestEi(sr dosa.SchemaRef) *dosa.EntityInfo {
 }
 
 type testCase struct {
-	encoder          encoding.Encoder
 	cachedEntities   []dosa.DomainObject
 	originRead       *expectArgs
 	originUpsert     *expectArgs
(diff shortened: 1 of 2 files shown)


connectors/cache/fallback_test.go
```diff
@@ -115,41 +104,44 @@ func TestUpsertCases(t *testing.T) {
...
 					"StrV":        "test value string",
-					"BoolV":       false,
+					"BoolVP":      &testBool,
 				}},
 			fallbackUpsert: &expectArgs{
 				values: map[string]dosa.FieldValue{
-					key:   []byte(`[{"an_uuid_key":"d1449c93-25b8-4032-920b-60471d91acc9"},{"strkey":"test key string"}]`),
-					value: []byte(`{"BoolV":false,"StrV":"test value string","an_uuid_key":"d1449c93-25b8-4032-920b-60471d91acc9","strkey":"test key string"}`),
+					// gob encoding of [{"an_uuid_key":"d1449c93-25b8-4032-920b-60471d91acc9"},{"strkey":"test key string"}]
+					key: []byte{13, 255, 131, 2, 1, 2, 255, 132, 0, 1, 255, 130, 0, 0, 14, 255, 129, 4, 1, 2, 255, 130, 0, 1, 12, 1, 16, 0, 0, 98, 255, 132, 0, 2, 1, 11, 97, 110, 95, 117, 117, 105, 100, 95, 107, 101, 121, 6, 115, 116, 114, 105, 110, 103, 12, 38, 0, 36, 100, 49, 52, 52, 57, 99, 57, 51, 45, 50, 53, 98, 56, 45, 52, 48, 51, 50, 45, 57, 50, 48, 98, 45, 54, 48, 52, 55, 49, 100, 57, 49, 97, 99, 99, 57, 1, 6, 115, 116, 114, 107, 101, 121, 6, 115, 116, 114, 105, 110, 103, 12, 17, 0, 15, 116, 101, 115, 116, 32, 107, 101, 121, 32, 115, 116, 114, 105, 110, 103},
```
rkuris: Please change this test. You should just encode, and then decode, and compare. Please do not rely on the exact byte output, as this can change due to a bug fix in the gob encoder.

We don't care what the actual bytes contain, we just want to be sure the encoder and decoder work correctly.

Another approach: this test can be redone if you allow for the encoder to be changed. The encoder can just produce a mock value, and you can check that it returns that.

connectors/cache/fallback_test.go
```diff
@@ -216,9 +208,11 @@ func TestReadCases(t *testing.T) {
 
 func createReadSuccessTestCase() testCase {
 	originResponse := map[string]dosa.FieldValue{"a": "b"}
+	// gob encoding of {"a":"b","primaryKey":"primaryValue"}
+	// note that gob encoding of maps is non-deterministic so the above map could be different bytes
+	gobEncodedResp := []byte{14, 255, 129, 4, 1, 2, 255, 130, 0, 1, 12, 1, 16, 0, 0, 52, 255, 130, 0, 2, 1, 97, 6, 115, 116, 114, 105, 110, 103, 12, 3, 0, 1, 98, 10, 112, 114, 105, 109, 97, 114, 121, 75, 101, 121, 6, 115, 116, 114, 105, 110, 103, 12, 14, 0, 12, 112, 114, 105, 109, 97, 114, 121, 86, 97, 108, 117, 101}
```
rkuris: this is another instance of where we don't care about the actual bytes.

PR #268, opened 2018-01-05
Overwrite canUpsertSchema in routing connector 

jaricftw:
This is needed to route the `canUpsertSchema` request to the correct connector.

### DIFF
--- a/entity.go
+++ b/entity.go
@@ -434,7 +434,7 @@ func (e *EntityDefinition) IsCompatible(e2 *EntityDefinition) error {
 			}
 
 			if !reflect.DeepEqual(index1, index2) {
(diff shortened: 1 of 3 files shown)


entity.go
```diff
@@ -434,7 +434,7 @@ func (e *EntityDefinition) IsCompatible(e2 *EntityDefinition) error {
 			}
 
 			if !reflect.DeepEqual(index1, index2) {
-				return errors.Errorf("index mismatch: (%v vs %v)", index1, index2)
+				return errors.Errorf("index %s mismatch: (%v vs %v)", name, index1, index2)
```
phliar: The error msg will be a little nicer if you use %q instead of %s

jaricftw: done.

PR #270, opened 2018-01-09
implement multiread for dosa client

junchaowu:


client.go
```diff
@@ -120,10 +120,9 @@ type Client interface {
 	// as a result of the read
 	Read(ctx context.Context, fieldsToRead []string, objectToRead DomainObject) error
 
-	// TODO: Coming in v2.1
 	// MultiRead fetches several rows by primary key. A list of fields can be
 	// specified. Use All() or nil for all fields.
-	// MultiRead(context.Context, []string, ...DomainObject) (MultiResult, error)
+	MultiRead(context.Context, []string, ...DomainObject) (MultiResult, error)
```
jaricftw: Can you add comments on how this API supports multiread, and what it does not support? For example, same entity, cross-entity, or cross-scope?

client.go
```diff
@@ -306,8 +305,58 @@ func (c *client) Read(ctx context.Context, fieldsToRead []string, entity DomainO
...
+	var listFieldValues []map[string]FieldValue
+	for _, entity := range entities {
+		ere, err := c.registrar.Find(entity)
+		if err != nil {
+			return nil, err
+		}
+
+		if re == nil {
+			re = ere
+		} else if re != ere {
```
jaricftw: For my own understanding, what are the different entity types?

client.go
```diff
@@ -306,8 +305,58 @@ func (c *client) Read(ctx context.Context, fieldsToRead []string, entity DomainO
...
+		return nil, err
+	}
+
+	multiResult := MultiResult{}
+	// map results to entity fields
+	for i, entity := range entities {
+		if results[i].Error != nil {
+			multiResult[entity] = results[i].Error
+			continue
+		}
```
jaricftw: I wonder if this cleaner to explicitly set the `entity` in the `multiResult` map if there is no error, i.e `multiResult[entity] = nil`. It seems easier for the caller to get the result, something like:
```
rs, err := range multiResult {
 if err != nil {
     // do something with rs
 }
} 
```

mocks/client.go
```diff
@@ -1,315 +1,142 @@
...
-// all copies or substantial portions of the Software.
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
+// Automatically generated by MockGen. DO NOT EDIT!
```
jaricftw: add Copyright

client.go
junchaowu: Basically this mean the multiread only support fetch one type of entity.

client.go
junchaowu: They should check the len(multiResult) to tell if there is any error or not.

jaricftw: Okay. I see that we use the same approach to retrieve result for the single Read API, so better keep it consistent for now, although I am not very convinced that this is the best way to do it.

PR #272, opened 2018-01-10
expose the IsDomainObject API

junchaowu:
Users want to have someway to tell whether a object is registered as DOSA object. Registrar has `find` method so that users can tell if an domainObject is registered or not.

### DIFF
--- a/client.go
+++ b/client.go
@@ -104,6 +104,9 @@ func ErrorIsAlreadyExists(err error) bool {
 
 // Client defines the methods to operate with DOSA entities
 type Client interface {
+	// GetRegistrar returns the registrar
+	GetRegistrar() Registrar
+
 	// Initialize must be called before any data operation
 	Initialize(ctx context.Context) error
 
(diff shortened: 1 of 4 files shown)


jaricftw:
Title should be updated. 

Can you add description on the context of this change? Seems the registar is passed in when creating a new client, I don't know why they still need it exposed.

client.go
```diff
@@ -231,6 +234,11 @@ func NewClient(reg Registrar, conn Connector) Client {
 	}
 }
 
+// GetRegistrar returns the registrar that is registered in the client
+func (c *client) GetRegistrar() Registrar {
+	return c.registrar
```
phliar: How unsafe is this?

Is it possible to return a copy?

junchaowu:
It is  true that the registrar is composed outside DOSA. However due to automation of initialization, the registrar is hidden by frameworks so that users don't have chance to hold the registrar.

client.go
junchaowu: Registrar is an Interface which provides readOnly methods. So it is safe to expose.

PR #273, opened 2018-01-10
Scope metadata

phliar:
Depends on https://github.com/uber/dosa-idl/pull/25


phliar:
Also updates yarpc-go to the latest (1.26.1), the prior version is too old to generate code.

connectors/yarpc/yarpc.go
```diff
@@ -633,8 +634,11 @@ func (c *Connector) CreateScope(ctx context.Context, scope string) error {
 
 // TruncateScope truncates all data in the scope specified
 func (c *Connector) TruncateScope(ctx context.Context, scope string) error {
+	// A better authn story is needed -- an evildoer could just craft a request with a
+	// bogus owner name and send it directly, bypassing this client.
 	request := &dosarpc.TruncateScopeRequest{
-		Name: &scope,
+		Name:      &scope,
+		Requester: dosa.GetUsername(),
```
jaricftw: how do you plan to authenticate this? Does the the requester of `TruncateScope` needs to be the same as the requester of `CreateScope`?  I suppose there could be multiple users accessing the same scope.

connectors/yarpc/yarpc.go
```diff
@@ -617,7 +617,8 @@ func (c *Connector) CheckSchemaStatus(ctx context.Context, scope, namePrefix str
 // CreateScope creates the scope specified
 func (c *Connector) CreateScope(ctx context.Context, scope string) error {
 	request := &dosarpc.CreateScopeRequest{
-		Name: &scope,
+		Name:      &scope,
+		Requester: dosa.GetUsername(),
```
jaricftw: Where do we store the requester information?

connectors/yarpc/yarpc.go
phliar: Same comment for all three methods -- right now it's easy to spoof, but after goldstar phase 2 (or whatever it's called) we should be able to do it right, with user-level authn/authz.

connectors/yarpc/yarpc.go
phliar: In DOSA! The gateway will store all scope metadata. The owning user is the only thing the server cannot compute so it must be provided by the client. If anyone can think of any other attributes it would be useful for the client to sent, let me know!

connectors/yarpc/yarpc.go
phliar: ... I guess we will have to figure out some sort of pluggable module thing for authz/authn, I don't goldstar is open-sourced.

phliar: Yes I think it makes sense that only the owner of a scope can drop or truncate it.

jaricftw: A scope should be owned by a team, not a particular user imo. Another option is to allow them specify the owner in the CLI, but that's also easy to spoof and we'll need to think about the migration of existing scopes. 

Is there a doc that I should look at :)

phliar: Hmmm team owned scopes are an interesting point. The only reasonable way to do it is to use LDAP groups, and confirm that the requesting user is in the requested group... and the server should be the one to verify that.

I'm going to add another field to the IDL, `owner`.  The user can use any string, the client won't verify it; the server will confirm that the owner is either the same as requester, or an LDAP group that the requester is part of.


Doc: https://docs.google.com/document/d/1sTehpUFV8qcbx_AlsXPmwlWyse3PvgnGcF59OEFPI4U/edit#

phliar: Updated the doc with team scopes

connectors/devnull/devnull.go
```diff
@@ -116,7 +116,7 @@ func (c *Connector) CheckSchemaStatus(ctx context.Context, scope, namePrefix str
 }
 
 // CreateScope returns success
-func (c *Connector) CreateScope(ctx context.Context, scope string) error {
+func (c *Connector) CreateScope(ctx context.Context, scope, owner string) error {
```
klnusbaum: nit: shouldn't `owner` actually be `_` to be consistent with the schema connector?

util.go
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
klnusbaum: nit: In accordence with my new-years resolution to get rid of files and packages named "util" or "lib", I suggest we rename this file to "user.go" (and "user_test.go" for it's accompanying test file).

connectors/yarpc/yarpc.go
jaricftw: Looks good to me

jaricftw:
LGTM overall, +1 that we don't merge to master for now. We plan to release the client soon as it is that has the CLI fix and multi-read API.

phliar:
> from a strict SemVer standpoint this is a breaking change and would require a major version bump

I'm starting to think that a major version bump is the right thing here, since the diff is also turning out to be a bit bigger than I'd initially thought.

client.go
```diff
@@ -708,8 +708,8 @@ func (ee *EntityErrors) Error() string {
 }
 
 // CreateScope creates a new scope
-func (c *adminClient) CreateScope(ctx context.Context, s string) error {
-	return c.connector.CreateScope(ctx, s)
+func (c *adminClient) CreateScope(ctx context.Context, s string, md *ScopeMetadata) error {
```
junchaowu: In this case, why not move the scope name inside the scopeMetaData. Anyway this is a breaking change, so we can combine them.

connector.go
```diff
@@ -77,6 +79,54 @@ type EntityInfo struct {
...
+// may be qualified by a prefix, as in "production.vsoffers".
+type ScopeMetadata struct {
+	Entity      `dosa:"primaryKey=(Name)" json:"-"`
+	Name        string     `json:"name"`
+	Owner       string     `json:"owner"` // Owning group name (or the same as Creator)
+	Type        int32      `json:"type"`  // Production, Staging, or Development
+	Flags       int64      `json:"flags"`
+	Version     int32      `json:"version"`
+	PrefixStr   string     `json:"prefix_str,omitempty"` // With ":" separators
+	Cluster     string     `json:"cluster,omitempty"`    // Host DB cluster
```
junchaowu: I am not sure about he Cluster. What's the scenario that if service migrates between dosa clusters.

connector.go
```diff
@@ -77,6 +79,54 @@ type EntityInfo struct {
...
+type StringSet map[string]struct{}
+
+// ScopeMetadata is metadata about a scope. (JSON tags to support MD setting CLI tools.) The scope
+// may be qualified by a prefix, as in "production.vsoffers".
+type ScopeMetadata struct {
+	Entity      `dosa:"primaryKey=(Name)" json:"-"`
+	Name        string     `json:"name"`
+	Owner       string     `json:"owner"` // Owning group name (or the same as Creator)
+	Type        int32      `json:"type"`  // Production, Staging, or Development
+	Flags       int64      `json:"flags"`
```
junchaowu: what's flags?

client.go
phliar: Good point....

connector.go
phliar: Migration is TBD -- in the normal case, a scope lives on exactly one cluster. For migration we'll have to set up some additional machinery which will have "Source" and "Destination" clusters.

connector.go
phliar: Initially I was going to add some parameters here but I changed the code and this is no longer needed.

I think in the future this will be useful, so my feeling is to leave it in. But I can remove it.

PR #276, opened 2018-01-12
Fix the memory connector findInsertionPoint bug

junchaowu:
The existing findInsertionPoint method doesn't return the right insertion point due to the comparison  returns the incorrect value which results the pagination in the range doesn't work correctly.

### DIFF
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -416,7 +416,7 @@ func (c *Connector) RemoveRange(_ context.Context, ei *dosa.EntityInfo, columnCo
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	partitionRange, err := c.findRange(ei, columnConditions, false)
+	partitionRange, _, err := c.findRange(ei, columnConditions, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


connectors/memory/memory.go
```diff
@@ -115,17 +115,14 @@ func findInsertionPoint(pk *dosa.PrimaryKey, data []map[string]dosa.FieldValue,
...
 func compareRows(pk *dosa.PrimaryKey, v1 map[string]dosa.FieldValue, v2 map[string]dosa.FieldValue) (cmp int8) {
 	keys := pk.ClusteringKeys
 	for _, key := range keys {
 		d1 := v1[key.Name]
 		d2 := v2[key.Name]
 		cmp = compareType(d1, d2)
-		if key.Descending {
-			cmp = -cmp
-		}
+
```
rkuris: I don't understand why this code was removed. I believe it's necessary for the case where you have a mix of ascending and descending columns in the clustering keys.

jaricftw: +1 for clarification. 

Even with a single clusterKey, we'll need it for for correctly comparing the target row with the existing rows. E.g. if the clustering key sorts the rows in DESC, we'll need to find the first row that's smaller than or equal to the target row's clustering key.

junchaowu: Updated the diff. There are some more issues in the memory conector

PR #277, opened 2018-01-18
Add ETL tag to Entity struct and pass that information through YARPC

jaricftw:
- Add an ETL tag to the entity struct for enabling data ingestion. Customer can enable ETL for a specific entity by setting `etl=on`;
- once the ETL flag is turn on, it cannot be turned off anymore. i.e. `etl=off` is NOT allowed;
- pass the `ETLState` information through YARPC
- fix a break change of error message in satori/go.uuid


connectors/yarpc/helpers.go
```diff
@@ -196,6 +196,29 @@ func PrimaryKeyToThrift(key *dosa.PrimaryKey) *dosarpc.PrimaryKey {
...
 
+// ETLStateToThrift convert the dosa ETLState to the thrift ETLState enum type
+func ETLStateToThrift(etl dosa.ETLState) dosarpc.ETLState {
+	switch etl {
+	case dosa.EtlOn:
+		return dosarpc.ETLStateOn
+	case dosa.EtlOff:
+		return dosarpc.ETLStateOff
+	default:
+		panic("bad ETL state")
```
junchaowu: print out the etl state number?

entity_parser.go
```diff
@@ -333,35 +335,72 @@ func parseNameTag(tag, defaultName string) (string, string, error) {
...
+	}
+	return fullETLTag, etlState, nil
+}
+
 // parseEntityTag function parses DOSA tag on the "Entity" field
-func parseEntityTag(structName, dosaAnnotation string) (string, *PrimaryKey, error) {
+func parseEntityTag(structName, dosaAnnotation string) (string, ETLState, *PrimaryKey, error) {
 	tag := dosaAnnotation
+
+	// find the ETL flag
```
junchaowu: since the ETL flag is optional , so let's parse the required tag such as primarykey first.

entity_parser_key_parser_test.go
```diff
@@ -526,8 +536,49 @@ func TestEntityParse(t *testing.T) {
 					},
 				},
 			},
+			ETL:   EtlOff,
 			Error: nil,
 		},
+		{
+			Tag:       "name=jj primaryKey=ok, etl=on",
```
junchaowu: are we choosing 'on/off' over 'enabled/disabled'? just confirm

entity_parser_key_parser_test.go
```diff
@@ -526,8 +536,49 @@ func TestEntityParse(t *testing.T) {
...
+			TableName: "jj",
+			PrimaryKey: &PrimaryKey{
+				PartitionKeys:  []string{"ok"},
+				ClusteringKeys: nil,
+			},
+			Error: nil,
+			ETL:   EtlOn,
+		},
+		{
+			Tag:       "name=jj primaryKey=ok etl=off",
```
junchaowu: Add tests for different cases of etl value. "ON/OFF" and "On/Off"

glide.lock
```diff
@@ -25,21 +25,17 @@ imports:
   - util/runes
   - util/strings
 - name: github.com/gocql/gocql
-  version: 33a5f3c1bcc2c421b3221c5858312afb141bf605
+  version: dd47639f787e8583e1a13e17e20a9f6c4332bc29
```
junchaowu: check the version for gocql

connectors/yarpc/helpers.go
jaricftw: done

jaricftw: done

entity_parser.go
jaricftw: Basically, I'd like to append the `etl` tag at the end, as it's more natural for users to add this to their new schema. However, due to the way we parse `primaryKey`, if we parse it first, it'll treat the entire substring (including `etl`) as primary key. Therefore I parse the tag backwards, i.e. `etl`->`primaryKey`->`name`.

We could change the parsing order by changing the way we parse primaryKey though, which is a bit out of the scope of this pull.

entity_parser_key_parser_test.go
jaricftw: yes. on/off seems more concise and accurate to me

entity_parser_key_parser_test.go
jaricftw: done

glide.lock
jaricftw: Have we finalized the version that breaks our client? 

I can pin the version in yaml for now, until we are safe to upgrade. Is there a ticket to track it?

jaricftw: I reverted the gocql version and pined it in yaml file

entity_parser.go
junchaowu: The reason to parse required tag first is to give user the most critical error first.  If the etl is part of the primarykey some how, we want to indicate the error of it.  If there is a field is called by the optional tag name such as 'name', 'etl' . we want to parse those first before we parse optional tag.

junchaowu: Being that said, we should include tests to have field name `etl` and `name`

jaricftw: done

PR #282, opened 2018-02-05
fix bad race conditions in memory connector

klnusbaum:
Currently, when reading, ranging, or scanning from the in-memory connector, the connector returns references to internal data (in the forms of slices and maps). This allows callers to read/write internal data without acquiring the appropriate locks inside the in-memory connector, resulting in race conditions.

By returning copies of the values stored within the in-memory connector, we can ensure the values returned to the caller have no references to the internal state within the in-memory connector and prevent data races.

### DIFF
--- a/connectors/memory/memory_test.go
+++ b/connectors/memory/memory_test.go
@@ -22,12 +22,12 @@ package memory
 
 import (
 	"context"
+	"io/ioutil"
+	"math/rand"
+	"sort"
+	"sync"
 	"testing"
 	"time"
-
-	"sort"
-
-	"math/rand"
 
 	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


connectors/memory/memory_test.go
```diff
@@ -1173,6 +1173,234 @@ func TestCompoundPartSecondaryIndex(t *testing.T) {
 	assert.EqualValues(t, readVals1["CreatedAt"], createdAt1)
 }
 
+func TestReadRaceNonClustered(t *testing.T) {
```
jaricftw: I am curious, will this test fail with `go test` in CI build, or we can only reproduce the race condition with `-race` flag on?

klnusbaum: Without the fix and without the `-race` flag on, this test will sometimes fail with just a `go test`, because one go-routine does the write before the read. If you then turn on -race, you'll see this problem pointed out by the race detector. With the problem now fixed though, this test will always pass, even if you don't run it with `-race`, because the actually problem is fixed.

PR #286, opened 2018-03-19
Add TTL feature

junchaowu:


entity_parser.go
```diff
@@ -365,20 +368,48 @@ func parseETLTag(tag string) (string, ETLState, error) {
...
+func parseTTLTag(tag string) (string, time.Duration, error) {
+	fullTTLTag := ""
+	ttlTag := ""
+	matches := ttlPattern0.FindStringSubmatch(tag)
+	if len(matches) == 2 {
+		fullTTLTag = matches[0]
+		ttlTag = matches[1]
+	}
+
+	if len(matches) == 0 {
```
klnusbaum: nit: I'd put this if block before the `len(matches) == 2` since it's the common case.

entity_parser.go
```diff
@@ -365,20 +368,48 @@ func parseETLTag(tag string) (string, ETLState, error) {
...
+	}
+
+	// filter out "trailing comma"
+	ttlTag = strings.TrimRight(ttlTag, " ,")
+	ttl, err := time.ParseDuration(ttlTag)
+	if err != nil {
+		return "", NoTTL(), err
+	}
+
+	if ttl < 0 {
```
klnusbaum: should we return an error in this case? Why would we want to silently fail on a negative TTL? Is there a reasonable usecase for negative TTLs? Seems indicative of an error.

entity_parser.go
```diff
@@ -365,20 +368,48 @@ func parseETLTag(tag string) (string, ETLState, error) {
...
 // parseEntityTag function parses DOSA tag on the "Entity" field
-func parseEntityTag(structName, dosaAnnotation string) (string, ETLState, *PrimaryKey, error) {
+func parseEntityTag(structName, dosaAnnotation string) (string, time.Duration, ETLState, *PrimaryKey, error) {
 	tag := dosaAnnotation
 
 	// find the primaryKey
 	matchs := primaryKeyPattern0.FindStringSubmatch(tag)
 	if len(matchs) != primaryKeyPattern0.NumSubexp()+1 {
-		return "", EtlOff, nil, fmt.Errorf("dosa.Entity on object %s with an invalid dosa struct tag %q", structName, tag)
+		return "", time.Duration(0), EtlOff, nil, fmt.Errorf("dosa.Entity on object %s with an invalid dosa struct tag %q", structName, tag)
```
klnusbaum: nit: let's be consistent with the rest of the code below and instead use `NoTTL()` here.

registry.go
```diff
@@ -30,54 +30,53 @@ import (
 // performing operations on an entity as well as helper methods for accessing
 // type data so that reflection can be minimized.
 type RegisteredEntity struct {
-	scope  string
-	prefix string
-	table  *Table
-	info   *EntityInfo
-	typ    reflect.Type // optimization to avoid doing repetitive reflect.TypeOf
+	scope     string
```
klnusbaum: this change seems completely unrelated to adding TTLs. Can you please move it to a separate PR?

ttl.go
```diff
@@ -20,9 +20,21 @@
...
 }
+
+// ValidateTTL returns whether the TTL is validated or not
+func ValidateTTL(ttl time.Duration) error {
+	if ttl < 1 * time.Second {
+		return errors.New("TTL is not allowed to set less than 1 second")
+	}
+
+	return nil
+}
```
klnusbaum: Nit: Missing linebreak?

PR #292, opened 2018-03-27
Always remove from cache on dosa upsert

tiffanycheng:
Safer to just remove from the cache, regardless of if the dosa call actually succeeded

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## v2.6.0 (unreleased)
 
 ## v2.5.3 (unreleased)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## v2.6.0 (unreleased)
 
 ## v2.5.3 (unreleased)
-- Invalidate fallback cache on upsert (#285)
+- Invalidate fallback cache on upsert (#285, #292)
```
phliar: v2.5.3 has already been released, let's put this diff into 2.5.4

PR #301, opened 2018-04-06
Implement multi upsert and multi remove methods in memory connector

mattgode:
This implements the missing multi upsert and multi remove methods in the memory connector. As noted, this does not use a lock when removing and upserting the set of entities. There is no such guarantee provided in the interface and thus we do not do it here.


klnusbaum:
Seems correct

PR #304, opened 2018-04-09
Implement MultiUpsert method in client and expose it via the Client interface

mattgode:
This is the last of 3 pull requests to complete functionality for multi upsert in the gateway. Multi remove is implemented in all other connectors but this diff does not implement/expose it in the client and defers that until the minor version release following the release of upsert.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v3.0.0 (unreleased)
+- Implement MultiUpsert method in client and expose it via the Client interface (#304)
```
phliar: I don't see any breaking changes here, what's the reason for wanting to call this 3.0?

tiffanycheng: i guess no reason. should i add a 2.7.0?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
-## v3.0.0 (unreleased)
+## v2.7.0 (unreleased)
```
junchaowu: we intend to release 3.0 as next release

PR #305, opened 2018-04-09
remove fqnIndex

klnusbaum:
The `fqnIndex` is never used in the `prefixedRegistrar`. It's dead code. Let's remove it.


klnusbaum:
As I explained in the description. this is dead code. It's a private map that is used by no other code in our entire codebase. Having dead code makes the library harder to understand. When reading the library and figuring out what it does, your mind now has to spend time figuring out and realizing that this bit of code actually doesn't matter and you should just ignore it. This seemed like a good bit of cleanup to remove some cognitive load from those trying to understand how our library works.

PR #309, opened 2018-04-24
Use Satori UUID and bump it to 1.2

phliar:
A lot of boilerplate and plumbing. Interesting bits:
 -  conversion to/from `gocql.uuid` to `satori/go.uuid.UUID` (via strings)
 - the `finder` needs to accept `uuid.UUID` as a permitted type


phliar:
Abandoning

PR #316, opened 2018-04-27
FindAll will not return error when no registered entities found

MaxJTRice:


junchaowu:
User Squash and Merge

PR #317, opened 2018-04-30
Glide cleanup

klnusbaum:
This is a large scale cleanup of our glide.yaml file. We remove a bunch of packages we don't need anymore, and clean up a little code related to packages that we actually didn't need. We also organize all of the packages we're bringing in that are purely related to build infrastructure (and not actually needed for building of the code).

### DIFF
--- a/.build/deps.mk
+++ b/.build/deps.mk
@@ -12,7 +12,7 @@ deps: libdeps
 	$(ECHO_V)go install ./vendor/github.com/mattn/goveralls
 	$(ECHO_V)go install ./vendor/github.com/go-playground/overalls
 	@$(call label,Installing golint...)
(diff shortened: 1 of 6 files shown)


.build/deps.mk
```diff
@@ -12,7 +12,7 @@ deps: libdeps
 	$(ECHO_V)go install ./vendor/github.com/mattn/goveralls
 	$(ECHO_V)go install ./vendor/github.com/go-playground/overalls
 	@$(call label,Installing golint...)
-	$(ECHO_V)go install ./vendor/github.com/golang/lint/golint
+	$(ECHO_V)go install ./vendor/golang.org/x/lint/golint
 	@$(call label,Installing errcheck...)
```
junchaowu: we can remove the  "get package" statement in .travis.yaml now

PR #322, opened 2018-05-03
Make all usages of $USER lowercase

eurbs:
Not all $USER variables are lowercase. This should cast all $USER env variables to lowercase to avoid broken commands for users that are reading READMEs or running bootstrapping scripts that use the dosacli without specifying a caller.

Further, many tests fail when contributing to the repo because of the reliance on $USER being lowercase. Instead of having to set USER=somethinglowercase before running `make`, this should simplify the dev flow.


connectors/cassandra/datastore_crud_test.go
```diff
@@ -45,8 +45,8 @@ func TestReadNotFound(t *testing.T) {
 
 func TestReadTimeout(t *testing.T) {
 	sut := GetTestConnector(t)
-	id := constructKeys(dosa.NewUUID())
```
klnusbaum: why do these things need to be changed?

eurbs: This was poor handling of a merge conflict. Good catch. It doesn't need to be changed.

eurbs:
Good to go. Thanks @klnusbaum !

- Datastore_crud_test.go was removed entirely in the v3 branch, so those weird changes are gone completely now.
- Rebased with v3 branch, all tests passing.

PR #323, opened 2018-05-03
Per-Entity dynamic TTL support

MaxJTRice:


registry.go
```diff
@@ -22,6 +22,7 @@ package dosa
```
phliar: This file has been renamed to registrar.go in v3....

registrar.go
```diff
@@ -53,11 +54,15 @@ func (e *RegisteredEntity) SetVersion(v int32) {
 
 // EntityInfo is a helper for accessing the registered entity's EntityInfo
 // instance which is required by clients to call connector methods.
-func (e *RegisteredEntity) EntityInfo() *EntityInfo {
+func (e *RegisteredEntity) EntityInfo(dynTTL *time.Duration) *EntityInfo {
```
junchaowu: Keep the original EntityInfo method interface. Change the TTL after you get the entityinfo instance since anyway the TTL is a public field.

PR #326, opened 2018-05-04
Client Shutdown

klnusbaum:
Currently, the client does not properly shutdown it's connectors. Let's have it do that.


client.go
```diff
@@ -182,6 +182,12 @@ type Client interface {
 	// To scan the next set of rows, modify the scanOp to provide
 	// the string returned as an Offset()
 	ScanEverything(ctx context.Context, scanOp *ScanOp) ([]DomainObject, string, error)
+
+	// Shutdown gracefully shutsdown the client, cleaning up any resources it may have
+	// allocated during it's usage. Shutdown should be called whenever the client
```
phliar: Two nits: 
 - shutsdown -> shuts down
 - "it's" == "it is", the possessive is "its"

client.go
```diff
@@ -216,6 +222,11 @@ type AdminClient interface {
 	TruncateScope(ctx context.Context, s string) error
 	// DropScope drops the scope and the data and schemas in the scope
 	DropScope(ctx context.Context, s string) error
+	// Shutdown gracefully shutsdown the client, cleaning up any resources it may have
```
phliar: Same comments

mocks/client.go
```diff
@@ -1,3 +1,23 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
phliar: 2018

PR #332, opened 2018-05-24
Add options for not invalidating cache when upsert

samUber:
Currently we invalidate the fallback cache every time we do an upset. Add options for some entities to skip invalidating.

### DIFF
--- a/connectors/cache/fallback.go
+++ b/connectors/cache/fallback.go
@@ -49,12 +49,31 @@ type rangeQuery struct {
 	Limit      int
 }
 
-// NewConnector creates a fallback cache connector
-func NewConnector(origin, fallback dosa.Connector, scope metrics.Scope, entities ...dosa.DomainObject) *Connector {
-	return newConnector(origin, fallback, scope, encoding.NewGobEncoder(), entities...)
+// Options returns a function that's being used for connector initialization
+type Options func(*Connector) error
+
+// WithSkipWriteInvalidateEntities provides the option for client to set the entites
+func WithSkipWriteInvalidateEntities(entities ...dosa.DomainObject) Options {
+	return func(c *Connector) error {
+		c.skipWriteInvalidateEntitiesMap = createSkipWriteCacheInvalidateSet(entities)
+		return nil
+	}
 }
 
-func newConnector(origin, fallback dosa.Connector, scope metrics.Scope, encoder encoding.Encoder, entities ...dosa.DomainObject) *Connector {
+// NewConnector creates a fallback cache connector
+func NewConnector(origin, fallback dosa.Connector, scope metrics.Scope, entities []dosa.DomainObject, options ...Options) *Connector {
+	c := newConnector(origin, fallback, scope, encoding.NewGobEncoder(), entities)
+	for _, option := range options {
+		err := option(c)
+		if err != nil {
+			// It's an no-op when getting an error now
+			continue
+		}
+	}
+	return c
+}
+
+func newConnector(origin, fallback dosa.Connector, scope metrics.Scope, encoder encoding.Encoder, entities []dosa.DomainObject) *Connector {
 	bc := base.Connector{Next: origin}
 	set := createCachedEntitiesSet(entities)
 	return &Connector{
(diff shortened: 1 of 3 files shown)


connectors/cache/fallback.go
```diff
@@ -49,12 +49,31 @@ type rangeQuery struct {
 	Limit      int
 }
 
+// Options returns a function that's being used for connector initialization
+type Options func(*Connector) error
+
+// WittSkipWriteInvalidateEntities provides the option for client to set the entites
+func WittSkipWriteInvalidateEntities(entities ...dosa.DomainObject) Options {
+	return func(c *Connector) error {
+		c.skipWriteInvalidateEntitiesMap = createSkipWriteCacheInvlidateSet(entities)
```
tiffanycheng: spelling

connectors/cache/fallback.go
```diff
@@ -49,12 +49,31 @@ type rangeQuery struct {
 	Limit      int
 }
 
+// Options returns a function that's being used for connector initialization
+type Options func(*Connector) error
+
+// WittSkipWriteInvalidateEntities provides the option for client to set the entites
+func WittSkipWriteInvalidateEntities(entities ...dosa.DomainObject) Options {
```
tiffanycheng: spelling

connectors/cache/fallback.go
```diff
@@ -393,13 +409,18 @@ func (c *Connector) cacheWrite(w func() error) error {
 	return nil
 }
 
+func (c *Connector) shouldSkipInvalidateCacheOnWrite(ei *dosa.EntityInfo) bool {
+	if c.skipWriteInvalidateEntitiesMap == nil {
```
tiffanycheng: don't need this check, it will just return false for nil map

connectors/cache/fallback_test.go
```diff
@@ -248,8 +258,17 @@ func TestUpsertCases(t *testing.T) {
 		},
 	}
 	for _, t := range testCases {
-		runTestCase(t)
+		runTestCaseWithInvalidateCache(t)
+	}
+	runTestCaseWithoutInvalidateCache := func(tc testCase) {
```
tiffanycheng: seems like your test should be failing. the test says it's still calling Remove on the fallback

connectors/cache/fallback.go
samUber: It only returns false because of the struct initialization.

connectors/cache/fallback_test.go
samUber: there were some issue with the mocking originally, I will help with fixing it.

connectors/cache/fallback.go
```diff
@@ -49,12 +49,31 @@ type rangeQuery struct {
...
+	return func(c *Connector) error {
+		c.skipWriteInvalidateEntitiesMap = createSkipWriteCacheInvalidateSet(entities)
+		return nil
+	}
+}
+
 // NewConnector creates a fallback cache connector
-func NewConnector(origin, fallback dosa.Connector, scope metrics.Scope, entities ...dosa.DomainObject) *Connector {
-	return newConnector(origin, fallback, scope, encoding.NewGobEncoder(), entities...)
+func NewConnector(origin, fallback dosa.Connector, scope metrics.Scope, entities []dosa.DomainObject, options ...Options) *Connector {
```
tiffanycheng: i don't think this function makes much sense if you're passing in Option that can only take a fallback Connector as an argument, but this function passes in a dosa.Connector interface.

connectors/cache/fallback_test.go
```diff
@@ -169,35 +169,66 @@ func (e *argsEncoder) Decode(byt []byte, int interface{}) error {
...
+
+	connector := NewConnector(mockOrigin, mockFallback, nil, cacheableEntities, failOption,
+		WithSkipWriteInvalidateEntities(cacheableEntities...))
+
+	assert.NotNil(t, connector)
+}
+
+// Test dosa upsert and the various behaviors of the fallback
+func TestUpsertCases(t *testing.T) {
+	runTestCaseWithInvalidateCache := func(tc testCase) {
```
tiffanycheng: can we move this to line 276 to group it with the other function?

samUber:
addressed comment

PR #349, opened 2018-07-19
Read-only scope metadata access for the CLI

phliar:
This diff adds readonly scope metadata access to the CLI. Two additional scope subcommands are added: 
 - `list` to print a list of all known scopes; and 
 - `show` to print the MD for the specified scope(s).

Other changes: `ClientProvider` is a bad name -- which "client"? Renamed to `AdminClientProvider` since it provides an admin client. Likewise, `provideYarpcClient` -- they're **all** yarpc clients!


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+
+## v3.1.1 (unreleased)
```
junchaowu: 3.2?

cmd/dosa/options.go
```diff
@@ -35,7 +35,13 @@ import (
 
 const _defServiceName = "dosa-gateway"
 
-var validNameRegex = regexp.MustCompile("^[a-z]+([a-z0-9]|[^-]-)*[^-]$")
+var (
+	// from YARPC: "must begin with a letter and consist only of dash-delimited
+	// lower-case ASCII alphanumeric words" -- we do this here because YARPC
+	// will panic if caller name is invalid.
+	validNameRegex = regexp.MustCompile("^[a-z]+([a-z0-9]|[^-]-)*[^-]$")
+	errmsg         = "callerName %s must begin with a letter and consist only of dash-separated lower-case ASCII alphanumeric words"
```
junchaowu: const?

mocks/client.go
```diff
@@ -1,23 +1,3 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
```
junchaowu: you need to setup the header hook.

cmd/dosa/scopemd.go
```diff
@@ -0,0 +1,127 @@
...
+		if md, err := c.getMetadata(client, scope); err != nil {
+			fmt.Printf("Could not read scope metadata for %q: %v\n", scope, err)
+		} else {
+			fmt.Printf("%+v\n", md)
+		}
+	}
+	return nil
+}
+
+func (c *ScopeShow) getMetadata(client dosa.Client, scope string) (*dosa.ScopeMetadata, error) {
```
junchaowu: Here it only takes scope as input without prefix. 
In dev, one scope has one name prefix
in prod, one scope has many prefixes. How does this method support that?

phliar: Currently, this only handles scopes, i.e. dev. Prefix support needs to be added, there is this comment in the defn for `ScopeMetadata`:

```
The scope may be qualified by a prefix, as in "production.vsoffers".
```

mocks/client.go
phliar: How do I do that?

CHANGELOG.md
phliar: This feature is for internal use only, I don't want to announce it! :)

cmd/dosa/scopemd.go
junchaowu: Add a comment on top of the method

cederber:
Can you split the boilerplate copyright updates out into a separate diff?

MaxJTRice:
Maybe it's better that we have a detailed description for this PR. And considering it's a CLI change we can attach some test examples.

phliar:
Xiaoxiao, made https://jira.uberinternal.com/browse/DOSA-1665 to document all this.

PR #351, opened 2018-07-23
Print results in tabulated format

jianli:


cmd/dosa/util.go
```diff
@@ -114,16 +117,35 @@ func strToFieldValue(t dosa.Type, s string) (dosa.FieldValue, error) {
 	}
 }
 
-func printResult(res []map[string]dosa.FieldValue) {
-	if len(res) == 0 {
-		fmt.Println("Error: not found")
-		return
+func printResults(results []map[string]dosa.FieldValue) error {
+	const padding = 3
```
MaxJTRice: Should this be a constant? If so maybe it's better to define outside this function (under the imports)?

cmd/dosa/util.go
```diff
@@ -114,16 +117,35 @@ func strToFieldValue(t dosa.Type, s string) (dosa.FieldValue, error) {
...
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, padding, ' ', tabwriter.Debug)
+	if len(results) == 0 {
+		return errors.New("Empty results")
```
MaxJTRice: You may need to fix a test for this error message :)

cmd/dosa/util.go
```diff
@@ -114,16 +117,35 @@ func strToFieldValue(t dosa.Type, s string) (dosa.FieldValue, error) {
...
-			fmt.Printf("%s: %v\n", field, reflect.Indirect(reflect.ValueOf(value)))
+	i := 0
+	width := len(results[0])
+	keys := make([]string, width)
+	for k := range results[0] {
+		keys[i] = k
+		i++
+	}
+	sort.Strings(keys)
+	_, err := fmt.Fprintln(w, strings.Join(keys, "\t"))
```
MaxJTRice: nit: you can have the same logic like this:

if _, err := fmt.Fprintln(w, strings.Join(keys, "\t")); err != nil {
    return errors.WithStack(err)
}

cmd/dosa/util.go
```diff
@@ -114,16 +117,35 @@ func strToFieldValue(t dosa.Type, s string) (dosa.FieldValue, error) {
...
+		return errors.WithStack(err)
+	}
+	values := make([]string, width)
+	for _, result := range results {
+		i := 0
+		for _, key := range keys {
+			values[i] = fmt.Sprintf("%v", reflect.Indirect(reflect.ValueOf(result[key])))
+			i++
+		}
+		_, err := fmt.Fprintln(w, strings.Join(values, "\t"))
```
MaxJTRice: nit: same as ^^

cmd/dosa/util.go
```diff
@@ -114,16 +117,35 @@ func strToFieldValue(t dosa.Type, s string) (dosa.FieldValue, error) {
...
+	}
+	sort.Strings(keys)
+	_, err := fmt.Fprintln(w, strings.Join(keys, "\t"))
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	values := make([]string, width)
+	for _, result := range results {
+		i := 0
+		for _, key := range keys {
```
MaxJTRice: nit: here maybe we can write like
'for idx, result := range results'
and get rid of i := 0 and i++?

cmd/dosa/util.go
```diff
@@ -114,16 +117,35 @@ func strToFieldValue(t dosa.Type, s string) (dosa.FieldValue, error) {
...
+	const padding = 3
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, padding, ' ', tabwriter.Debug)
+	if len(results) == 0 {
+		return errors.New("Empty results")
 	}
-	fmt.Println("Results:")
-	for _, fieldValues := range res {
-		for field, value := range fieldValues {
-			fmt.Printf("%s: %v\n", field, reflect.Indirect(reflect.ValueOf(value)))
+	i := 0
```
MaxJTRice: nit: maybe it's better to have a meaningful parameter name rather than 'i'? :)

cmd/dosa/query.go
```diff
@@ -99,7 +99,10 @@ func (c *QueryCmd) doQueryOp(f func(ShellQueryClient, context.Context, []*queryO
 		return err
 	}
 
-	printResult(results)
+	err = printResults(results)
```
junchaowu: put it in one line

jianli:
@rkuris I've added unit tests as well as escaping for whitespace characters.

Our use case is quickly getting DOSA rows for visual inspection (in the same way that you would query any standard SQL interface) so we don't need HTML or JSON output.

cmd/dosa/util.go
```diff
@@ -114,16 +117,35 @@ func strToFieldValue(t dosa.Type, s string) (dosa.FieldValue, error) {
...
+		return errors.New("Empty results")
 	}
-	fmt.Println("Results:")
-	for _, fieldValues := range res {
-		for field, value := range fieldValues {
-			fmt.Printf("%s: %v\n", field, reflect.Indirect(reflect.ValueOf(value)))
+	idx := 0
+	width := len(results[0])
+	keys := make([]string, width)
+	for k := range results[0] {
```
MaxJTRice: Found 1 issue from my tests here when there's empty (null) value for the results:
e.g. the table has three columns (c1, c2, c3), we insert two rows: 

row1: (r1c1, nil, r1c3)
row2: (r2c1, r2c2, r2c3)

Two possibilities:
1. If row1 is results[0], then the CLI will only show columns c1 and c3
2. If row2 is results[0], then the CLI will show columns c1, c2 and c3, but row1's c2 value will be: <invalid reflect.Value>

Suggestions:
1. For printResults function, you can pass the list of fields to show (add fields to input signature). You can get the fields from: https://github.com/uber-go/dosa/blob/master/cmd/dosa/query.go#L92
2. We check if result[key] is valid. If not, we print 'nil' for it like cqlsh does.

PR #358, opened 2018-08-28
PoC: ScopeMetadata and PrefixMetadata

phliar:
Two structs and an interface: Getters and Setters, oh my!

### DIFF
--- a/prefix_metadata.go
+++ b/prefix_metadata.go
@@ -1,0 +1,217 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


prefix_metadata.go
```diff
@@ -0,0 +1,221 @@
...
+func (md *PrefixMetadata) GetFlags() int64 {
+	return md.Flags
+}
+
+// GetVersion is the version of the prefix.
+func (md *PrefixMetadata) GetVersion() int32 {
+	return md.Version
+}
+
+// GetEntityStr is the encoded list of all prefixes.
```
phliar: More useful methods would be `AddEntity()`, `HasEntity() bool`, and `GetEntities() []string`....

phliar: (and same for prefixes)

PR #359, opened 2018-08-28
Fix the FIXME check

phliar:
Don't search for a substring `fixme`, look for it as a word


phliar:
Old behaviour:

```
Checking for unresolved FIXMEs...
metadata.go:// Metadata is an interface abstracting both ScopeMetadata and PrefixMetadata.
prefix_metadata.go:// PrefixMetadata is metadata about a prefix in production (note: we do not track prefixes for dev scopes).
prefix_metadata.go:type PrefixMetadata struct {
prefix_metadata.go:func (md *PrefixMetadata) String() string {
...
```

Issue #343, opened 2018-07-11
Replace package satori/go.uuid


josselin-c:
This issue has been opened automaticaly because we think you are using satori/uuid in your code. Unfortunately, satori/uuid has a critical defect where it doesn't generate random UUIDV4s satori/go.uuid#73 and should be replaced as the maintainer has been unresponsive for 5 months with the bug still open. Alternative packages you should consider: https://github.com/google/uuid https://github.com/pborman/uuid

rkuris:
See also https://github.com/gofrs/uuid

rkuris:
See https://github.com/uber-go/dosa/pull/361

MaerF0x0:
Reviving the dead here, but do not see  https://github.com/gofrs/uuid as it's "A UUID package originally forked from github.com/satori/go.uuid"

PR #361, opened 2018-08-29
Switch to gofrs/uuid

rkuris:
The satori/go.uuid package has some serious known problems, particularly
with the generation of V4 UUIDs. It's also no longer maintained. The
recommended replacement is gofrs/uuid.

Additionally, the gofrs team accepted the change I made to the uuid
package about two years ago and released 3.1.0 which includes a feature
needed by the DOSA memory connector. This also removes that workaround.

### DIFF
--- a/connectors/memory/memory.go
+++ b/connectors/memory/memory.go
@@ -24,13 +24,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
-	"encoding/binary"
 	"sort"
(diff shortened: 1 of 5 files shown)


rkuris:
Fixes https://github.com/uber-go/dosa/issues/343

klnusbaum:
seems reasonable to me. @phliar are you ok with this?

PR #364, opened 2018-09-11
Cleanup EnsureValidRangeConditions

phliar:
Breaking out of a nested loop using a label is barely better than a GOTO, and is a code smell.


phliar:
https://jira.uberinternal.com/browse/DOSA-1734

PR #369, opened 2018-10-16
test: replace missed satori/go.uuid (#361)

twilly:
961abe5 swapped out `satori/go.uuid` with `gofrs/uuid` but missed this one spot in a routing connector test.

This pull request is broken into three parts. The first is formatting issues (`go fmt` from v1.10). The second is the fix for a missed `satori/go.uuid` replacement. The last is updating dependencies, which drops `satori/go.uuid` from glide.lock, but because of glide I'm also including updates of other dependent libraries. Details are in each commit message.

### DIFF
--- a/cmd/dosa/main.go
+++ b/cmd/dosa/main.go
@@ -23,10 +23,11 @@ package main
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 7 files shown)


twilly:
Closing this PR since it's been many years without comment.

PR #374, opened 2019-05-15
Add a 'columns' tag for Index definitions that defines which fields should be denormalized into the index

skywhat:
The index table is built on selecting all the fields in the entity. It’s costly to read/write with large payload. Using columns tag to make it flexible,  which could reduce the payload when build the materialized view.

This pull is to extend the annotation. Users could choose to use **columns=(field1, field2, field3)** to limit the columns required to build the index table.

### DIFF
--- a/entity.go
+++ b/entity.go
@@ -166,13 +166,19 @@ func (cd *ColumnDefinition) Clone() *ColumnDefinition {
 
 // IndexDefinition stores information about a DOSA entity's index
 type IndexDefinition struct {
-	Key *PrimaryKey
+	Key     *PrimaryKey
+	Columns []string
 }
 
 // Clone returns a deep copy of IndexDefinition
 func (id *IndexDefinition) Clone() *IndexDefinition {
+	columns := make([]string, len(id.Columns))
+	for i, c := range id.Columns {
+		columns[i] = c
+	}
 	return &IndexDefinition{
-		Key: id.Key.Clone(),
+		Key:     id.Key.Clone(),
+		Columns: columns,
 	}
 }
 
--- a/entity.go
+++ b/entity.go
@@ -173,9 +173,7 @@ type IndexDefinition struct {
 // Clone returns a deep copy of IndexDefinition
 func (id *IndexDefinition) Clone() *IndexDefinition {
 	columns := make([]string, len(id.Columns))
-	for i, c := range id.Columns {
-		columns[i] = c
-	}
+	copy(columns, id.Columns)
 	return &IndexDefinition{
 		Key:     id.Key.Clone(),
 		Columns: columns,
--- a/entity_parser.go
+++ b/entity_parser.go
@@ -49,7 +49,7 @@ var (
 
 	namePattern0 = regexp.MustCompile(`name\s*=\s*(\S*)`)
 
-	columnsPattern0 = regexp.MustCompile(`columns\s*=\s*\(([^\(\)]+)\)`)
+	columnsPattern = regexp.MustCompile(`columns\s*=\s*\(([^\(\)]+)\)`)
 
 	etlPattern0 = regexp.MustCompile(`etl\s*=\s*(\S*)`)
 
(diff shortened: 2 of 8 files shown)


phliar:
Are you sure this is not a breaking change?

entity_parser_key_parser_test.go
```diff
@@ -749,6 +750,7 @@ func TestIndexParse(t *testing.T) {
 				ClusteringKeys: nil,
 			},
 			InputIndexName: "SearchByKey",
+			Columns:        []string{},
```
phliar: What happens if you don't put `Columns` into this literal? You may be able to leave it off.

entity.go
```diff
@@ -166,13 +166,19 @@ func (cd *ColumnDefinition) Clone() *ColumnDefinition {
...
 type IndexDefinition struct {
-	Key *PrimaryKey
+	Key     *PrimaryKey
+	Columns []string
 }
 
 // Clone returns a deep copy of IndexDefinition
 func (id *IndexDefinition) Clone() *IndexDefinition {
+	columns := make([]string, len(id.Columns))
+	for i, c := range id.Columns {
```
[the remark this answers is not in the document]
skywhat: Good idea!

entity_parser.go
```diff
@@ -49,6 +49,8 @@ var (
 
 	namePattern0 = regexp.MustCompile(`name\s*=\s*(\S*)`)
 
+	columnsPattern0 = regexp.MustCompile(`columns\s*=\s*\(([^\(\)]+)\)`)
```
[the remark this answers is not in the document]
skywhat: no idea but followed others' coding style. easily misled

entity_parser.go
```diff
@@ -338,6 +348,30 @@ func parseNameTag(tag, defaultName string) (string, string, error) {
 	return fullNameTag, name, nil
 }
 
+// parseColumnsTag functions parse DOSA "columns" tag
```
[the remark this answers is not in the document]
skywhat: update

entity_parser_key_parser_test.go
skywhat: I start to understand what's difference between `[]string{}` and `[]string(nil)` in golang. Why still no idea why golang is designed in this way :(

entity_parser.go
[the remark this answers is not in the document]
phliar: You can leave that in. At some point we need to re-implement this whole file, it's terrible.

entity_parser_key_parser_test.go
phliar: Yes, the two things are different. However the interesting question is: are they different enough to matter? That is why I suggested trying it out -- if you remove the lines, does it still compile and work as expected?

phliar:
A better title for this diff would be "Add a 'columns' tag for Index definitions that defines which fields should be denormalized into the index."

phliar:
> Are you sure this is not a breaking change?

For completeness: we analyzed the situation, and found that adding fields to Go structs **_is_** a breaking change. We looked at all the apps that use DOSA and decided that in our case it is acceptable, and we don't need to treat it as a breaking change.

entity_parser.go
```diff
@@ -338,6 +348,30 @@ func parseNameTag(tag, defaultName string) (string, string, error) {
 	return fullNameTag, name, nil
 }
 
+// parseColumnsTag parses "columns" tag of dosa.Index in the entity
+// returns full string of columns and a slice of column fields.
```
phliar: Use this:
```
// parseColumnsTag parses the "columns" tag of a dosa.Index in the entity. It returns
// the matched section of the tag string and a list of the selected fields.
```

skywhat: matched section! I like this.

entity_parser_key_parser_test.go
skywhat: The reason I could remove these lines is I changed the default value of `Columns` from `[]string{}` to `[]string(nil)`

skywhat:
> A better title for this diff would be "Add a 'columns' tag for Index definitions that defines which fields should be denormalized into the index."

skywhat:
> A better title for this diff would be "Add a 'columns' tag for Index definitions that defines which fields should be denormalized into the index."

Thanks for being so helpful. It looks much better. I even learned what's denormalization. so professional!

phliar:
Code looks good. One last thing: add a line to `CHANGELOG.md` in the unreleased section at the top. I suggest something simple like "Add optional 'columns' tag to Index definitions."

PR #376, opened 2019-05-24
update code related to 'columns' tag for index definition

skywhat:
1. update dependency packages such as latest dosa-idl
2. cover corner cases
   1)  fields in columns tag must refer to a column
   2) a column cannot be used twice
3. update dosa schema dump feature due to flexible secondary index
4. update Thrift related function in /yarpc

### DIFF
--- a/schema/cql/cql.go
+++ b/schema/cql/cql.go
@@ -55,15 +55,20 @@ func typeMap(t dosa.Type) string {
 	return "unknown"
 }
 
+func isSelectAllCreatingView(columns []string) bool {
+	return len(columns) == 0
+}
+
 // precompile the template for create table
 var cqlCreateTableTemplate = template.Must(template.
 	New("cqlCreateTable").
 	Funcs(map[string]interface{}{"typeMap": typeMap}).
 	Funcs(map[string]interface{}{"uniqueKey": uniqueKey}).
+	Funcs(map[string]interface{}{"isSelectAllCreatingView": isSelectAllCreatingView}).
 	Parse(`create table "{{.Name}}" ({{range .Columns}}"{{- .Name -}}" {{ typeMap .Type -}}, {{end}}primary key {{ .Key }});
 {{- range $name, $indexdef := .Indexes }}
 create materialized view "{{- $name -}}" as
-  select * from "{{- $.Name -}}"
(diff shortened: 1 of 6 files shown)


schema/cql/cql.go
```diff
@@ -55,15 +55,20 @@ func typeMap(t dosa.Type) string {
...
 var cqlCreateTableTemplate = template.Must(template.
 	New("cqlCreateTable").
 	Funcs(map[string]interface{}{"typeMap": typeMap}).
 	Funcs(map[string]interface{}{"uniqueKey": uniqueKey}).
+	Funcs(map[string]interface{}{"isSelectAllCreatingView": isSelectAllCreatingView}).
 	Parse(`create table "{{.Name}}" ({{range .Columns}}"{{- .Name -}}" {{ typeMap .Type -}}, {{end}}primary key {{ .Key }});
 {{- range $name, $indexdef := .Indexes }}
 create materialized view "{{- $name -}}" as
-  select * from "{{- $.Name -}}"
+  select {{if isSelectAllCreatingView $indexdef.Columns}}*{{else}}{{range $colnum, $col := $indexdef.Columns}}{{if $colnum}}, {{end}}"{{$col}}"{{end}}{{end}} from "{{- $.Name -}}"
```
phliar: What is the test that tells us there is no error in this line?

schema/cql/cql_test.go
```diff
@@ -71,7 +80,18 @@ create materialized view "i2" as
   where "int64type" is not null
   primary key (int64type, booltype ASC);`,
 		},
-		// TODO: Add more test cases
```
phliar: We no longer need to add more test cases?

phliar:
Are these changes related, or separate? If they are separate changes we should not combine them in one diff.

skywhat:
> Are these changes related, or separate? If they are separate changes we should not combine them in one diff.
I think they could be treated as related cause these changes are needed due to the introduction of our new flexible secondary index feature.

schema/cql/cql_test.go
skywhat: kk. put it back

schema/cql/cql.go
skywhat: https://github.com/uber-go/dosa/blob/index/schema/cql/cql_test.go#L84

PR #379, opened 2019-05-24
add columns support to materialized view creation using Schema dump

skywhat:


schema/cql/cql.go
```diff
@@ -55,15 +55,20 @@ func typeMap(t dosa.Type) string {
...
 var cqlCreateTableTemplate = template.Must(template.
 	New("cqlCreateTable").
 	Funcs(map[string]interface{}{"typeMap": typeMap}).
 	Funcs(map[string]interface{}{"uniqueKey": uniqueKey}).
+	Funcs(map[string]interface{}{"isSelectAllCreatingView": isSelectAllCreatingView}).
 	Parse(`create table "{{.Name}}" ({{range .Columns}}"{{- .Name -}}" {{ typeMap .Type -}}, {{end}}primary key {{ .Key }});
 {{- range $name, $indexdef := .Indexes }}
 create materialized view "{{- $name -}}" as
-  select * from "{{- $.Name -}}"
+  select {{if isSelectAllCreatingView $indexdef.Columns}}*{{else}}{{range $colnum, $col := $indexdef.Columns}}{{if $colnum}}, {{end}}"{{$col}}"{{end}}{{end}} from "{{- $.Name -}}"
```
phliar: Same question as the last time we saw this: how do we know that this is the correct CQL statement?

schema/cql/cql.go
```diff
@@ -55,15 +55,30 @@ func typeMap(t dosa.Type) string {
 	return "unknown"
 }
 
+func selectFieldsInCreatingView(columns []string) string {
+	if len(columns) == 0 {
+		return "*"
+	}
+	fields := ""
+	for colnum, col := range columns {
```
phliar: You can use `strings.Join` here, https://golang.org/pkg/strings/#Join

schema/cql/cql_test.go
```diff
@@ -82,6 +82,30 @@ create materialized view "i2" as
...
 }
 
+func TestSelectFieldsInCreatingView(t *testing.T) {
+	data := []struct {
+		Columns   []string
+		Statement string
+	}{
+		{
+			Columns:   []string{"foo", "bar", "hello"},
+			Statement: "\"foo\", \"bar\", \"hello\"",
```
phliar: You should use "raw strings" so you don't need so much quoting.
```
Statement: `"foo", "bar", "hello"`,
```

phliar:
Please update the diff title also, something like `add columns support to materialized view creation`.

If this is going to be the final diff, you will need a line in the changelog also.

schema/cql/cql.go
skywhat: Cool! strings magic!

PR #381, opened 2019-06-04
Rename CLI command "schema check" to "schema upsertable"

phliar:
Calling it `schema check` leads to confusion with the API method `CheckSchema`. If `CheckSchema` fails at service startup, people think they can debug the situation with `dosa schema check` -- which is not true.


skywhat:
Looks like `schema_test.go` also needs corresponding change.
https://github.com/uber-go/dosa/blob/master/cmd/dosa/schema_test.go#L293

PR #385, opened 2019-06-11
use append but not copy to clone

skywhat:
```
var s = make([]int, 3)
n := copy(s, []int{0, 1, 2, 3}) // n == 3, s == []int{0, 1, 2}
```
`copy(dst, src []T) int`
The number of elements copied is the minimum of len(src) and len(dst).
https://golang.org/ref/spec#Appending_and_copying_slices

In our use case, copy is not the right way to clone.

**make test**
**go test -v --run TestClone**

### DIFF
--- a/entity.go
+++ b/entity.go
@@ -173,7 +173,9 @@ type IndexDefinition struct {
 // Clone returns a deep copy of IndexDefinition
 func (id *IndexDefinition) Clone() *IndexDefinition {
 	var columns []string
-	copy(columns, id.Columns)
+	for _, c := range id.Columns {
+		columns = append(columns, c)
+	}
 	return &IndexDefinition{
 		Key:     id.Key.Clone(),
 		Columns: columns,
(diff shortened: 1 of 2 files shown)


entity.go
```diff
@@ -173,7 +173,9 @@ type IndexDefinition struct {
 // Clone returns a deep copy of IndexDefinition
 func (id *IndexDefinition) Clone() *IndexDefinition {
 	var columns []string
```
phliar: We could fix this by instead saying `columns := make([]string, len(id.Columns)` right?

skywhat: tried to use make to construct the string array before when I was debugging the string(nil) and string{} issue. The key point is that. when the upsert get into Schema service. it's stored as nil but not []string{}, so, after the first upsert, when u check upsertable again, index incompatible happened as well. And I believe it's also the way why all the previous Clone are using  nil and append.
it's super tricky, and hard to debug. didn't find any comment and explanation. jump into the trap is the only way to know..

PR #387, opened 2019-06-21
Add a 'defunct' tag for Index definitions that define the index is defunct or not

skywhat:
Cassandra doesn’t support changing columns of the materialized view. Therefore It introduces the concern that how to handle the existing index table since the goal of the feature is to reduce the large partition.
The current solution is to educate users to mark their old index table as `defunct` and then create a new index table with limited columns. 
The next time users use range query, the gateway would skip the lookup in defunct index table.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## v3.4.7 (unreleased)
- - Nothing changed yet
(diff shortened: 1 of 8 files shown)


phliar:
The title can be shorted to `Add a 'defunct' tag for Index definitions`

PR #391, opened 2019-07-12
 Add String() methods to all components of the routing connector

phliar:


skywhat:
Could I get more context why we need this change?

skywhat:
> Could I get more context why we need this change?
I got it. I saw this task. T3411623

skywhat:
would we get this new version released? if so, gateway imports new one and integration tests get fixed

phliar:
Not just about that particular task: in general all types should have a good `String()` method that will help in debugging. (Sometimes a plain `"%+v"` is enough, but not always.)

phliar:
> would we get this new version released? if so, gateway imports new one and integration tests get fixed

I don't need this released immediately. If you have a change that you want to include in a gateway build, we can release them together.

PR #392, opened 2019-07-15
Tests for the routing connector

phliar:


connectors/routing/config_test.go
```diff
@@ -27,42 +27,41 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-// TestBasicConfig test the basic yaml file conversion
-func TestBasicConfig(t *testing.T) {
-	yamlFile := `
+var yamlFile = `
 routers:
-# routers structure looks like:
```
awaniim: why remove these comments?

skywhat:
> The spec for the routing connector:
> 
> ```
> // Config for the routing connector is a "case statement" of scope names, and each entry is a list
> // of assigments "pattern" -> engine-name.
> //
> // Example:
> //
> // routers:
> // - production:
> //     serviceA: cassandra
> //     serviceX: schemaless
> //     *: dosa
> // - development:
> //     *: dosa_dev
> //     serviceB: cassandra
> //     serviceX: schemaless
> // - ebook:
> //     '*': ebook
> //     apple.*: ebook
> //     ebook-store: ebook
> // - default:
> //     sless_*: schemaless
> //     "*": dosa_dev
> //
> // A pattern is not a regular expression: only prefixes may be specified (i.e. trailing "*").
> // Literal strings (no "*") sort before a pattern.
> // "default" is an alias for "*" and sorts after everything else.
> ```

where do u get the routing grammar, do we need to add the comment in this `routing` file?

phliar:
BTW this is for information only, no need to accept.....

phliar:
Not for commit

phliar:
> where do u get the routing grammar, do we need to add the comment in this `routing` file?

It is not documented anywhere, I figured this out by reading the code. (Obviously the final diff will include the spec that we are trying to parse.)

PR #393, opened 2019-07-18
Routing connector fixes

phliar:
- sorting for []Rule was completely bogus
- tests for the routing connector were woefully inadequate
- lots of weirdness between "default" and "*"
- simplified errors
- made most things package-local instead of exported

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## v3.4.8 (unreleased)
  - Nothing changed yet
 
-## v3.4.7 (2019-07-20)
+## v3.4.7 (2019-07-22)
  - Add String() method to all components of the routing connector
  - Fix many bugs in the routing connector
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v3.4.8 (unreleased)
+ - Nothing changed yet
+
 ## v3.4.7 (unreleased)
  - Add String() method to all components of the routing connector
  - Fix many bugs in the routing connector
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## v3.4.7 (unreleased)
- - Nothing changed yet
+ - Add String() method to all components of the routing connector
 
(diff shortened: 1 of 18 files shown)


connectors/routing/routing_config.go
```diff
@@ -21,57 +21,165 @@
...
+		scopeIsPattern:  scopeIsPattern,
+		scopePattern:    scopePat,
+		prefixIsPattern: prefixIsPattern,
+		prefixPattern:   prefixPat,
+		canonScope:      nScope,
+		canonPfx:        nPrefix,
+	}, nil
+}
+
+// RouteTo (bad name!) says whether or not this Rule can handle the given scope:prefix.
```
skywhat: why don't we change the name if it's a bad name

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## v3.4.7 (unreleased)
- - Nothing changed yet
+ - Add String() method to all components of the routing connector
```
skywhat: but this diff is not limited to adding String() now, right?

names.go
```diff
@@ -27,66 +27,52 @@ import (
...
 	namePrefixRegex = regexp.MustCompile("^[a-z_][a-z0-9_.]{0,31}$")
+	nameRegex       = regexp.MustCompile("^[a-z_][a-z0-9_]{0,31}$")
 )
 
-// IsValidNamePrefix checks if a name prefix (a namespace prefixed to all DOSA entity type names)
-// conforms the following rules:
-// 1. name prefix starts with [a-z_]
-// 2. the rest of name prefix can contain only [a-z0-9_.]
-// 3. the length of name prefix must be greater than 0 and less than or equal to 32
+// IsValidNamePrefix checks if a name prefix.
```
skywhat: is this comment completed?

connectors/routing/routing_config.go
phliar: A diff is not ready for reviewing until I add reviewers.....

names.go
phliar: :)

phliar:
- made almost everything private
- added test to confirm that `footer < foo*`

PR #394, opened 2019-07-23
Parameterizing yarpc transport

ben-mays:
This change allows us to configure the dosa client transport as part of the yarpc connector config. 

YARPC does the heavy lifting, allowing us to use any supported transport (`tchannel`, `http`, and `grpc` are currently supported). The existing client behavior always uses tchannel, which is quickly becoming deprecated. 

I added an option (`--transport`) to the cli to allow cli users to specify their transport. By default, we fallback to tchannel. I chose not to expose the grpc transport, as it is still experimental and I do not have a grpc server to test against yet- the transport built is valid according to the grpc yarpc documentation.

### DIFF
--- a/cmd/dosa/main.go
+++ b/cmd/dosa/main.go
@@ -70,7 +70,8 @@ func (b BuildInfo) Execute(args []string) error {
 type GlobalOptions struct {
 	Host        string     `long:"host" default:"127.0.0.1" description:"The hostname or IP for the gateway."`
 	Port        string     `short:"p" long:"port" default:"21300" description:"The hostname or IP for the gateway."`
-	ServiceName string     `short:"s" long:"service" default:"dosa-gateway" description:"The TChannel service name for the gateway."`
+	ServiceName string     `short:"s" long:"service" default:"dosa-gateway" description:"The service name for the gateway."`
+	Transport   string     `long:"transport" default:"tchannel" description:"The transport to use when connecting to the gateway. Valid options: tchannel, http."`
 	CallerName  callerFlag `long:"caller" default:"dosacli-$USER" description:"The RPC Caller name."`
 	Timeout     timeFlag   `long:"timeout" default:"60s" description:"The timeout for gateway requests. E.g., 100ms, 0.5s, 1s. If no unit is specified, milliseconds are assumed."`
 	Version     bool       `long:"version" description:"Display version info"`
--- a/connectors/yarpc/yarpc.go
+++ b/connectors/yarpc/yarpc.go
@@ -24,14 +24,22 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
+
+	"crypto/tls"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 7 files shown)


glide.yaml
```diff
@@ -30,27 +30,31 @@ import:
   - transport/tchannel
 - package: gopkg.in/yaml.v2
   version: ^2.2.1
+- package: google.golang.org/grpc
+  version: ^1.22.0
+- package: github.com/prometheus/procfs
```
ben-mays: This package is somehow in the dep graph and does not have a major version. It's included as part of the prometheus client.

cmd/dosa/main.go
```diff
@@ -70,7 +70,8 @@ func (b BuildInfo) Execute(args []string) error {
 type GlobalOptions struct {
 	Host        string     `long:"host" default:"127.0.0.1" description:"The hostname or IP for the gateway."`
 	Port        string     `short:"p" long:"port" default:"21300" description:"The hostname or IP for the gateway."`
-	ServiceName string     `short:"s" long:"service" default:"dosa-gateway" description:"The TChannel service name for the gateway."`
+	ServiceName string     `short:"s" long:"service" default:"dosa-gateway" description:"The service name for the gateway."`
```
phliar: Can other transports be addressed by name? I'd think it is not relevant for HTTP... (perhaps I'm wrong)

connectors/yarpc/yarpc.go
```diff
@@ -73,14 +85,66 @@ type Connector struct {
 	headers    map[string]string
 }
 
+func buildYARPCConfig(config Config) (rpc.Config, error) {
+	ycfg := rpc.Config{Name: config.CallerName}
+	hostPort := fmt.Sprintf("%s:%s", config.Host, config.Port)
+	switch strings.ToLower(config.Transport) {
+	case tchannelTransport:
+		// this looks wrong, BUT since it's a uni-directional tchannel
```
phliar: Thanks for preserving the comment!

cmd/dosa/main.go
ben-mays: In this case we're conflating `CallerName` with `ServiceName` which in my mind implies the destination service. YARPC passes the destination service and destination procedure in all requests, regardless of transport.

After poking around, this may just be a idiosyncrasy specific to tchannel, where we require a service name on outbound creation: 

- Comment from abg: https://github.com/yarpc/yarpc-go/blob/ff8d19f1accd1b94e829b698d205a7258f82ddc9/transport/tchannel/channel_outbound.go#L118
- `NewChannelTransport` comments: https://github.com/yarpc/yarpc-go/blob/6151af6af236bde9aa512c58d57661265ef28522/transport/tchannel/channel_transport.go#L49

I'm not familiar with tchannels design but I'm guessing these are just named bidirectional channels and service name is the key.

skywhat:
Want to understand more. ```The existing client behavior always uses tchannel, which is quickly becoming deprecated.```
so if the tchannel deprecation is done, we will remove this option, am I right? 
And is there any announcement that the deprecation is coming?

PR #395, opened 2019-07-25
Cleanup String methods, and do not use reflect.DeepEqual to compare entities

phliar:
Instead of `reflect.DeepEqual` use sensible comparisons.
Also made `newRule` package-private to remove a lint check.

### DIFF
--- a/entity.go
+++ b/entity.go
@@ -22,10 +22,8 @@ package dosa
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
-
-	"reflect"
-
 	"time"
 
 	"github.com/pkg/errors"
@@ -55,6 +53,15 @@ func (ck ClusteringKey) String() string {
 	return ck.Name + " ASC"
 }
 
+// Format a key's ClusteringKeys.
+func formatClusteringKeys(keys []*ClusteringKey) string {
+	pieces := make([]string, len(keys))
+	for index, ck := range keys {
+		pieces[index] = ck.String()
+	}
+	return strings.Join(pieces, ", ")
+}
+
 // PrimaryKey stores information about partition keys and clustering keys
 type PrimaryKey struct {
 	PartitionKeys  []string
@@ -70,7 +77,6 @@ func (pk PrimaryKey) Clone() *PrimaryKey {
 		for i, k := range pk.PartitionKeys {
 			npk.PartitionKeys[i] = k
 		}
-
 	}
 
 	if pk.ClusteringKeys != nil {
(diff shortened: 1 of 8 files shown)


phliar:
No tests have been changed, and everything passes.

entity.go
```diff
@@ -55,6 +53,15 @@ func (ck ClusteringKey) String() string {
 	return ck.Name + " ASC"
 }
 
+// Format a key's ClusteringKeys.
+func formatClusteringKeys(keys []*ClusteringKey) string {
+	pieces := make([]string, len(keys))
+	for index, ck := range keys {
```
ben-mays: Does order matter? Seems like this method will return different results for the same keyset.

ben-mays: Re-phrasing my last comment: do we assume these are always sorted?

entity.go
```diff
@@ -182,6 +206,20 @@ func (id *IndexDefinition) Clone() *IndexDefinition {
 	}
 }
 
+func (id *IndexDefinition) String() string {
+	return fmt.Sprintf("%+v", *id)
```
ben-mays: This seems a bit shaky to me, if the `fmt` package makes a breaking change we could be in trouble.

ben-mays:
I'm a bit lost on how this affects dosa's operations, can you provide context on the existing behavior and the change's impact?

entity.go
```diff
@@ -182,6 +206,20 @@ func (id *IndexDefinition) Clone() *IndexDefinition {
...
+}
+
+func (id *IndexDefinition) equal(other *IndexDefinition) error {
```
skywhat: so instead of using `DeepEqual`, we will our own equal, is it risky which might introduce unexpected unequal thing.
I don't get it what's reason we are giving up DeepEqual. If we want more error info from `DeepEqual`, just use `%+v` rather than `%v` is good enough, that's how I debug before...

entity.go
phliar: > I'm a bit lost on how this affects dosa's operations, can you provide context on the existing behavior and the change's impact?

Absolutely no changes that any application will see. The only changes are in the details of the errors -- which have been carefully chosen so that they still use the same words that the tests check for, but for a human reading them there will be more detail about exactly what the mismatch is.

Order matters for clustering keys, but I'm not sure how that connects to this code so I wonder if I understand your question.

entity.go
phliar: Why would `fmt` make a breaking change? We would be in trouble if _any_ of the standard packages made a breaking change....

entity.go
phliar: I don't understand this. What does DeepEqual buy us?

entity.go
```diff
@@ -86,6 +92,45 @@ func (pk PrimaryKey) Clone() *PrimaryKey {
...
+func (pk PrimaryKey) clusteringKeysString() string {
+	if pk.ClusteringKeys == nil || len(pk.ClusteringKeys) == 0 {
+		return ""
+	}
+	return ", " + formatClusteringKeys(pk.ClusteringKeys)
+}
+
+func (pk PrimaryKey) equal(other *PrimaryKey) error {
+	if !stringSliceEqual(pk.PartitionKeys, other.PartitionKeys) {
+		return errors.Errorf("partition key mismatch: (%v vs %v)", pk.PartitionKeys, other.PartitionKeys)
```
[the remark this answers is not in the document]
phliar: The type of `pk.PartitionKeys` is `[]string`, and there is no difference between `%v` and `%+v`.

entity.go
phliar: To elaborate further: the reason we should not use DeepEqual here is because we don't necessarily want to compare each member of a struct to check equality. 

For instance, an object may have an internal unique ID field, or some pre-compiled cached stuff, or ... We want to make sure the user did not screw up when the schema change was made, so the code should only check the things that came from the user. We should not check fields whose values may change between executions of the program.

entity.go
ben-mays: > Order matters for clustering keys, but I'm not sure how that connects to this code so I wonder if I understand your question.

Got it, I was missing the context around the order here.

entity.go
ben-mays: I agree that it's an unlikely concern but something to consider. From the pkg docs, it appears to format `%v` in the following way:

```
struct:             {field0 field1 ...}
array, slice:       [elem0 elem1 ...]
maps:               map[key1:value1 key2:value2 ...]
pointer to above:   &{}, &[], &map[]
```

For `ColumnDefinitions`, this will include the `Tags` field - which may have a random ordering. If we compare two equivalent `ColumnDefinitions#String` outputs we may have different results right. I don't full understand how the String and Equals methods are consumed so I'll defer to your judgement.

entity.go
skywhat: I am convinced and understand more that DeepEqual has done too much. Thanks for the detailed explanation.

entity.go
phliar: That is an excellent point.  I don't know if it makes any actual difference but I'm not a fan of non-deterministic code so I'm going to fix that.

phliar: Changed `ColumnDefinition.String()` to be deterministic.

PR #396, opened 2019-08-16
modify the test cases using space as the delimiter between tags

skywhat:


entity_parser_key_parser_test.go
```diff
@@ -416,17 +416,6 @@ func TestEntityParse(t *testing.T) {
 		TTL        time.Duration
 		Error      error
 	}{
-		{
-			Tag:       "name=jj primaryKey=ok",
```
skywhat: delete but not modify cause the next one is the same

PR #398, opened 2019-08-22
refactor entity parser

skywhat:


entity_parser.go
```diff
@@ -40,141 +39,10 @@ const (
...
-	namePattern0 = regexp.MustCompile(`name\s*=\s*(\S*)`)
-
-	columnsPattern = regexp.MustCompile(`columns\s*=\s*\(([^\(\)]+)\)`)
-
-	etlPattern0 = regexp.MustCompile(`etl\s*=\s*(\S*)`)
-
-	ttlPattern0 = regexp.MustCompile(`ttl\s*=\s*(\S*)`)
-
 	indexType = reflect.TypeOf((*Index)(nil)).Elem()
+	tagsKey   = []string{"primaryKey", "key", "name", "columns", "etl", "ttl"}
```
ateliersakura: I see the same strings are repeated in the below code, which can be error-prone. How about defining them as variables like:

primaryKeyTag = "primaryKey";

and use "primaryKeyTag" in the remaining places so that compiler will tell you whether or not you are spelling correctly.

entity_parser.go
```diff
@@ -241,242 +109,402 @@ func TableFromInstance(object DomainObject) (*Table, error) {
...
+		return "", nil, nil, errors.Wrapf(err, "index %s has invalid columns tag %q", indexName, dosaAnnotation)
 	}
-	tag = strings.Replace(tag, fullColumnsTag, "", 1)
 
-	tag = strings.TrimSpace(tag)
-	if tag != "" {
-		return "", nil, nil, fmt.Errorf("index field %s with an invalid dosa index tag: %s", indexName, tag)
+	var key *PrimaryKey
+	if val, ok := tagsMap["key"]; ok {
+		key, err = parsePrimaryKey(val)
```
ateliersakura: Given that we use the same function for paring primary keys and index keys, how about calling it "parseKey"

entity_parser.go
```diff
@@ -241,242 +109,402 @@ func TableFromInstance(object DomainObject) (*Table, error) {
...
-func parseNameTag(tag, defaultName string) (string, string, error) {
-	fullNameTag := ""
-	name := defaultName
-
-	matches := namePattern0.FindStringSubmatch(tag)
-	if len(matches) == 2 {
-		fullNameTag = matches[0]
-		name = matches[1]
+// parseSinglePrimaryKey func parses the single parimary key of DOSA object
+func parseSinglePrimaryKey(s string) (*PrimaryKey, error) {
```
ateliersakura: parimary => primary

entity_parser.go
```diff
@@ -241,242 +109,402 @@ func TableFromInstance(object DomainObject) (*Table, error) {
...
+	s = strings.TrimSpace(s)
+	spaceIdx := strings.Index(s, " ")
```
ateliersakura: Will we disallow spaces between key fields like "xxx, yyy"?

Will this impact existing users?

entity_parser.go
```diff
@@ -241,242 +109,402 @@ func TableFromInstance(object DomainObject) (*Table, error) {
...
-	var fields []string
-	matches := columnsPattern.FindStringSubmatch(tag)
-	if len(matches) == 2 {
-		fullColumnsTag = matches[0]
-		fields = strings.Split(matches[1], ",")
-	}
+// parseCompoundPartitionKeys func parses the compund partition key of DOSA object
+func parseCompoundPartitionKeys(s string) ([]string, error) {
+	s = strings.TrimSpace(strings.Trim(s, "()"))
+	fields := strings.Split(s, ",")
```
ateliersakura: do we need to check for the existence of spaces like the function above?

entity_parser.go
```diff
@@ -241,242 +109,402 @@ func TableFromInstance(object DomainObject) (*Table, error) {
...
+		ckFields = append(ckFields, field)
 	}
 
-	if len(matches) == 0 {
-		return "", EtlOff, nil
+	cks := []*ClusteringKey{}
+	for _, ckField := range ckFields {
+		fields := strings.Fields(ckField)
+		if len(fields) == 0 {
+			continue
```
ateliersakura: Should this be an error?

PR #407, opened 2019-10-14
Cleanup some CLI error messages

phliar:
This will also be v3.4.10.


client.go
```diff
@@ -294,7 +294,7 @@ func (c *client) Initialize(ctx context.Context) error {
 	// fetch latest version for all registered entities, assume order is preserved
 	version, err := c.connector.CheckSchema(ctx, c.registrar.Scope(), c.registrar.NamePrefix(), eds)
 	if err != nil {
-		return errors.Wrap(err, "CheckSchema failed")
+		return errors.Wrap(err, "schema on the server is inccompatible with the code")
```
skywhat: incompatible. typo

phliar: Earlier rev, already fixed

PR #409, opened 2019-10-17
improve the error message if name is invalid

skywhat:


names.go
```diff
@@ -42,19 +42,23 @@ var (
 	nameRegex       = regexp.MustCompile("^[a-z_][a-z0-9_]{0,31}$")
 )
 
+// DosaNamingRule DOSA Name Rule
+const DosaNamingRule = `DOSA valid names must start with a letter or underscore,
```
phliar: I don't think we should embed a newline in this message...

names.go
```diff
@@ -42,19 +42,23 @@ var (
...
+and may contain letters, digits, and underscores, and must not be longer than 32 characters.`
+
 // IsValidNamePrefix checks if a name prefix is valid.
 func IsValidNamePrefix(namePrefix string) error {
 	normalized := strings.ToLower(strings.TrimSpace(namePrefix))
 	if !namePrefixRegex.MatchString(normalized) {
-		return errors.Errorf("invalid name-prefix '%s'", namePrefix)
+		return errors.Errorf("invalid name-prefix '%s' %s", namePrefix, DosaNamingRule)
```
phliar: Insert a colon: `"invalid name-prefix '%s': %s"`

skywhat: Looks elegant!

PR #410, opened 2019-11-01
Add the concept of a table alias for an index

phliar:
Docstore does not implement index aliases yet.


phliar:
Docstore has not implemented index aliases yet; an index's table name may be different from the name of the index. (In the Docstore case the index gets a version number added to the end of the name, as in `initindex_v42`.)

skywhat:
So would we remove it someday after docstore has the index aliases?

phliar:
> So would we remove it someday after docstore has the index aliases?

Yes, unless something else comes up! :)

PR #414, opened 2019-12-14
Add support for nil partition key for secondary indexes in memory connector

aniketpant:
Why:
Cassandra supports secondary indexes where the partition key can be
nil. This results in the index not being generated until the value is
set. The test client which uses a memory connector should support the
same behavior.

-

This change addresses the need by:
Add a `nil` check in `partitionKeyBuilder` which is capable of handling
different value types. In these cases, we simply skip over the value
instead of encoding a `nil` value which results in a panic inside the
gob encoder.

-


connectors/memory/memory.go
```diff
@@ -93,6 +93,9 @@ func partitionKeyBuilder(pk *dosa.PrimaryKey, values map[string]dosa.FieldValue)
 	var encodedKey []byte
 	for _, k := range pk.PartitionKeys {
 		if v, ok := values[k]; ok {
+			if isNilInterface(v) {
```
phliar: Why not just `v == nil`?

aniketpant: I tired doing that at first but seems like it's one of those interface issues where an interface holding a nil value will be non-nil. An alternative to `isNilInterface` would be using reflection.

```
func isNilValue(i interface{}) bool {
    return i == nil || reflect.ValueOf(i).IsNil()
}
```

Example: https://play.golang.org/p/j9SqfEgGe4B

Ref: https://tour.golang.org/methods/12
> Note that an interface value that holds a nil concrete value is itself non-nil.

aniketpant: @phliar Do you know if there's an alternative here? See my comment on the issue I'm experiencing and why we need to cast to the interface.

rkuris: Your proposed PR is a more efficient way of doing it than adding a reflect call. A golang interface contains both a [Type](https://golang.org/pkg/reflect/#Type) and a [Value](https://golang.org/pkg/reflect/#Value), and you only want to see if the Value is nil. 
At this point in the code, the Type will never be nil, and so checking if the interface is nil will always be false.

There's a longer discussion of this here: https://groups.google.com/forum/#!topic/golang-nuts/wnH302gBa4I/discussion

connectors/memory/memory.go
```diff
@@ -714,6 +717,28 @@ func getStartingPoint(ei *dosa.EntityInfo, token string) (start string, startPar
 	return start, startPartKey, nil
 }
 
+func isNilInterface(v dosa.FieldValue) bool {
```
rkuris: isNilInterface could use a unit test

aniketpant: You're right. Will add one.

connectors/memory/memory.go
aniketpant: Yeah, this totally makes sense now.

I wanted to avoid using reflect here primarily because we are always going to deal with known types here.

PR #425, opened 2020-01-31
Make thrift decode of EntityDefinition respect column ordering.

phliar:
Should have included this in the previous diff (facepalm)....


connectors/yarpc/helpers.go
```diff
@@ -278,14 +278,17 @@ func FromThriftToPrimaryKey(key *dosarpc.PrimaryKey) *dosa.PrimaryKey {
 // FromThriftToEntityDefinition converts the RPC EntityDefinition to client EntityDefinition
 func FromThriftToEntityDefinition(ed *dosarpc.EntityDefinition) *dosa.EntityDefinition {
 	fields := make([]*dosa.ColumnDefinition, len(ed.FieldDescs))
-	i := 0
-	for k, v := range ed.FieldDescs {
+	colOrder := ed.ColumnOrder
+	if colOrder == nil {
+		colOrder = getRPCEntityColumns(ed)
+	}
+	for i, colName := range ed.ColumnOrder {
```
skywhat: so `ed.ColumnOrder` has the same value as `colOrder`, am I right?

phliar: Yes, this part of the code handles the case when an app uses an older client (before 3.4.15) -- if that happens with schema upserts, Docstore will throw an error saying the order is bad.

PR #443, opened 2020-05-28
Add cache configuration per endpoint in fallback cache

ViddyCat:
1. We are using existing 'methodName' context already being set and used in EPG
2. Create initialization function to be called as contructor options
3. New configuration defaults to NO-OP if config is not set or empty else is enforced

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## v3.4.27 (unreleased)
+## v3.4.26 (unreleased)
  - Add cache configuration per endpoint in fallback cache
 
 ## v3.4.26 (unreleased)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,6 @@
 ## v3.4.26 (unreleased)
  - Add cache configuration per endpoint in fallback cache
 
-## v3.4.26 (unreleased)
- - Nothing changed yet.
-
 ## v3.4.25 (2020-05-05)
  - Add the RateLimit error.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v3.4.27 (unreleased)
+ - Add cache configuration per endpoint in fallback cache
+
 ## v3.4.26 (unreleased)
  - Nothing changed yet.
 
--- a/connectors/cache/fallback.go
+++ b/connectors/cache/fallback.go
@@ -61,24 +61,24 @@ func WithSkipWriteInvalidateEntities(entities ...dosa.DomainObject) Options {
 }
 
 type ContextKey string
+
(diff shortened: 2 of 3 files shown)


connectors/cache/fallback.go
```diff
@@ -60,6 +60,22 @@ func WithSkipWriteInvalidateEntities(entities ...dosa.DomainObject) Options {
...
+var (
+	// ContextEndpoint allows users to pass in calling endpoint name
+	ContextEndpoint string = "endpoint"
+)
+
+// SetCacheableEndpoints sets cacheable endpoints
+func SetCacheableEndpoints(endpoints ...string) Options {
+	return func(c *Connector) error {
+		for _, endpoint := range endpoints {
+			c.cacheableEndpoints[endpoint] = true
```
phliar, line 73: I would expect an array named `cacheableEndpoints` to contain endpoints; not the status of endpoints. This should be called something like `cacheableEndpointStatus` (although that is a very long name).

I'm also not a fan of bool vars, an enum with descriptive names is more readable. I won't insist on that, though -- I'll leave it between you and your conscience.

phliar: Also, please add a line to the CHANGELOG describing the change.

ViddyCat: Done!

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v3.4.27 (unreleased)
+ - Add cache configuration per endpoint in fallback cache
```
phliar, line 4: This should go into 2.4.26, which is the  unreleased version.

connectors/cache/fallback.go
```diff
@@ -60,6 +60,24 @@ func WithSkipWriteInvalidateEntities(entities ...dosa.DomainObject) Options {
 	}
 }
 
+// Context key used by SetCacheableEndpoints
+var (
+	// ContextEndpoint allows users to pass in calling endpoint name
+	ContextEndpoint string = "endpoint"
+    // EndpointActiveStatus marks the endpoint as active
+    EndpointActiveStatus bool = true
```
phliar, line 68: Do these vars really need to be exported? They're not used anywhere else, right?

ViddyCat: Nope, fixed!

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v3.4.26 (unreleased)
+ - Add cache configuration per endpoint in fallback cache
```
phliar, line 4: There can't be two sections named `v3.4.26`! :)

ViddyCat: Dang...sorry!

ViddyCat: Fixed

PR #454, opened 2021-03-08
Use go modules

phliar:
And remove glide stuff.


awaniim (approved):
why the version jump from 29 to 31?

phliar:
> why the version jump from 29 to 31?

We didn't update it to 30 when releasing 3.4.30 :facepalm:!

PR #456, opened 2022-07-19
Return actual error message during client intialization

vvksh:
Currently the error messages are hard-coded and hide the actual error which makes debugging hard.


awaniim (approved):

ramiyer (approved):
Lgtm

